PR #58, opened 2022-08-15
feat: asset manager enhancements

### LINKED ISSUE #57: feat: asset manager enhancements
- [x] Asset Manager should be able to call a function on the pair to trigger a report
- [x] hook on Pair's mint and burn such that it triggers arbitrary logic on the AssetManager on those events

xenide:
## Motivation
1. asset should be notified of liquidity addition and removal, so that it can move assets to an optimal allocation 

## Solution
1. please review the draft approach I have proposed over here, and see if it is the direction we want to go
   - in terms of the targeted thresholds for managed / non-managed, we might have need a mapping for each pair 
2. for `burn`, we need to check if we even have enough token for redemption. If not enough, then we need to redeem some tokens from the asset manager to fulfill the request


xenide:
in terms of optimizations, I was thinking we can collapse the two callback functions into one. 

Besides that, we need to take of the case where burn requires more tokens than available in reserve. But I think we'll take care of this in vexchange/v3-core#98 in another PR

OliverNChalk (changes requested):
Looks good and clean, some minor comments

OliverNChalk (changes requested):
Can we do a call to decide on two callbacks vs one?

OliverNChalk (approved):

PR #66, opened 2022-08-19
test: add testMintFee_WhenRampingA

andersonaguiar:


xenide (commented):
yes this time you have changed the A by shifting the time, good job.

However, if there's no swaps, the liquidity and therefore `D` did not grow. So we cannot tell if the implementation has succeeded. We need to perform several swaps (can put them in a loop) to grow the liquidity. (see point 2 of my prev PR comment). 

Another way to do it is, create another stable pair, so we have `StablePair1` and `StablePair2`. Both of them experience the same liquidity changes (growth in D). But one has a ramping A, the other has a static A. Then we do a `mint` / `burn` / `mintFee` again. Then we see if the platform fee collector received the same amount of LP tokens as platform fees.

test/__fixtures/BaseTest.sol
```diff
@@ -27,6 +27,7 @@ abstract contract BaseTest is Test {
     constructor()
     {
         // set shared variables
+        _factory.set(keccak256("ConstantProductPair::platformFeeTo"), bytes32(uint256(uint160(_platformFeeTo))));
         _factory.set(keccak256("ConstantProductPair::swapFee"), bytes32(uint256(30)));
         _factory.set(keccak256("ConstantProductPair::platformFee"), bytes32(uint256(2500)));
         _factory.set(keccak256("ConstantProductPair::defaultRecoverer"), bytes32(uint256(uint160(_recoverer))));
```
OliverNChalk, line 33: Thoughts on refactoring this to use library? Would be good to lock in/test the library approach

OliverNChalk (approved):

test/__fixtures/BaseTest.sol
xenide: actually you didn't have to review this PR cuz it was the other one that mattered. I was going to close this PR. 

By library you mean like managing the setting + casting of these variables?

OliverNChalk: yes

xenide: will fix in a separate PR as this PR is old and no longer relevant

xenide:
closing as this has been addressed in #86

PR #75, opened 2022-09-07
feat: asset mgmt base class

### LINKED ISSUE #73: refactor(pair): move asset management functions into a shared base class

xenide:
## Motivation

- abstract asset management logic into a base class to avoid repetition and drift 

## Solution

- please see my approach in `AssetManagedPairTest::testSetManager` for the table test. If this approach is okay, I will implement it for all tests in the file
- there might still be 1/2 tests that work on specific curves, for example `testManageReserves_KStillHolds` and `testSync` as the StablePair doesn't have a `sync` function (yet)
- for `IAssetManager::adjustManagement()`, is there. reason we are using `int224` in the mock implementation instead of `int256` or `int112`?


src/asset-management/AssetManagedPair.sol
```diff
@@ -0,0 +1,164 @@
...
+
+    modifier onlyFactory() {
+        require(msg.sender == address(factory), "AMP: FORBIDDEN");
+        _;
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                ASSET MANAGEMENT
+
+    Asset management is supported via a two -way interface. The pool is able to
```
OliverNChalk:
```suggestion
    Asset management is supported via a two-way interface. The pool is able to
```

src/asset-management/AssetManagedPair.sol
```diff
@@ -0,0 +1,164 @@
...
+    ask the current asset manager for the latest view of the balances. In turn
+    the asset manager can move assets in/ out of the pool. This section
```
OliverNChalk:
```suggestion
    the asset manager can move assets in/out of the pool. This section
```

src/asset-management/AssetManagedPair.sol
```diff
@@ -0,0 +1,164 @@
...
+    implements the pool - side of the equation. The manager's side is abstracted
```
OliverNChalk:
```suggestion
    implements the pool side of the equation. The manager's side is abstracted
```

src/curve/stable/StablePair.sol
```diff
@@ -200,7 +186,9 @@ contract StablePair is UniswapV2ERC20, ReentrancyGuard, IAssetManagedPair {
     /// @dev Mints LP tokens - should be called via the router after transferring tokens.
     /// The router must ensure that sufficient LP tokens are minted by using the return value.
     function mint(address to) public nonReentrant returns (uint256 liquidity) {
-        (uint256 _reserve0, uint256 _reserve1, ) = _getReserves();
+        _syncManaged();
+
+        (uint112 _reserve0, uint112 _reserve1, ) = _getReserves();
```
OliverNChalk, line 190: Was this a bug or why did the ordering change (maybe just for cleanness)?

src/interfaces/IAssetManagedPair.sol
```diff
@@ -1,8 +1,12 @@
 pragma solidity 0.8.13;
 
+import "src/interfaces/IAssetManager.sol";
```
OliverNChalk: Can we do explicit imports, file level imports considered unnecessary

src/interfaces/IAssetManager.sol
```diff
@@ -6,5 +6,6 @@ import { IAssetManagedPair } from "src/interfaces/IAssetManagedPair.sol";
 interface IAssetManager
 {
     function getBalance(IAssetManagedPair owner, address token) external returns (uint112 tokenBalance);
+    function adjustManagement(IAssetManagedPair pair, int256 amount0Change, int256 amount1Change) external;
```
OliverNChalk: This function is not used by the pairs right? What was the motive for adding it?

test/unit/AssetManagedPair.t.sol
```diff
@@ -0,0 +1,194 @@
...
+        // assert
+        assertEq(address(aPair.assetManager()), address(_manager));
+    }
+
+    function testSetManager() external
+    {
+        for (uint i = 0; i < _pairs.length; ++i) {
+            _setManager(_pairs[i]);
+        }
+    }
```
OliverNChalk: Try the modifier approach until foundry has table tests implemented:

https://github.com/foundry-rs/foundry/issues/858#issuecomment-1236870228

OliverNChalk (changes requested):

src/asset-management/AssetManagedPair.sol
xenide: fixed

src/asset-management/AssetManagedPair.sol
xenide: fixed

src/asset-management/AssetManagedPair.sol
xenide: fixed

src/curve/stable/StablePair.sol
xenide: so early today when I was doing the refactoring I realized that I did not call `_syncManaged` for `mint` and `burn`, hence the addition

src/interfaces/IAssetManagedPair.sol
xenide: fixed

src/interfaces/IAssetManager.sol
xenide: correct, not used by the pairs. 

My thought was that we should define all the external functions called by pairs and owner. If we only need the functions pair needs, then we can remove it. Please advise, thanks

test/unit/AssetManagedPair.t.sol
xenide: okay I tried it, it works, but it doesn't work if we require a clean slate for the test.

In other words, the state persists in between the different pairs. 

This is the same problem as the approach I had previously actually. I ran tests to see for myself that both approaches have the same problem

src/curve/stable/StablePair.sol
OliverNChalk: In general, I'm a big believer in the approach:

1. Find bug
2. Write unit test that fails, proving the bug
3. Fix the bug
4. If the unit test now passes you've proven the fix and the bug

This is key to quality & security over time, the alternative is hasty patches and vulnerability over time. I implemented this approach religiously in `vault` and it saved me from some silly regression bugs a few times.

src/interfaces/IConstantProductPair.sol
```diff
@@ -1,10 +1,8 @@
 pragma solidity 0.8.13;
 
-import { IAssetManager } from "src/interfaces/IAssetManager.sol";
-import { IAssetManagedPair } from "src/interfaces/IAssetManagedPair.sol";
 import { GenericFactory } from "src/GenericFactory.sol";
 
-interface IConstantProductPair is IAssetManagedPair {
+interface IConstantProductPair {
```
OliverNChalk, line 5: Why did we remove the inheritance here? Does doing it here mean the discrete pairs are not compatible?

test/unit/AssetManagedPair.t.sol
OliverNChalk: Damn, that sucks. Well until there a table tests in foundry we have two options. **1.** duplicate the test cases per pair in each pairs test file. or **2**, create an asset management test file that uses this approach to test both pairs:
```
function testSetManager_CaseA()
{
    _testSetManager_CaseA(_constantPair);
    _testSetManager_CaseA(_stablePair);
}

function _testSetManager_CaseA(IAssetManagedPair aPair) private
{
    // run test
}
```

Let me know which approach you're leaning towards, I lean towards option 2

xenide: okay I tried the second approach, it's the same problem as using the modifier. Cuz at the end of the day, forge creates a fresh instance for each public `test*` function there is. 

So I think it's only option 1 for now

OliverNChalk: You can use `vm.revertTo` to deal with this (could even use the modifier approach that snapshots, runs test, reverts:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.15;

import "test/__helpers/BaseTest.sol";
import {MintableERC20} from "test/__mocks/MintableERC20.sol";

import {IUniswapV2Factory} from "uni-v2-solc8/interfaces/IUniswapV2Factory.sol";
import {IFindFee} from "src/find-fee/IFindFee.sol";
import {FindFee} from "src/find-fee/FindFee.sol";

struct TestCase {
    string network;
    address target;
    bool isFactory;
    uint256 fee;
}

struct Fork {
    bool created;
    uint256 forkId;
}

contract FindFeeTest is BaseTest {
    IFindFee private _findFee;
    TestCase[] private _testCases;

    mapping(string => Fork) private _forks;

    function setUp() external {
        _testCases.push(
            TestCase({network: "ethereum", target: 0x111111111117dC0aa78b770fA6A738034120C302, isFactory: false, fee: 0})
        );
        _testCases.push(
            TestCase({network: "ethereum", target: 0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f, isFactory: true, fee: 30})
        );
        _testCases.push(
            TestCase({network: "ethereum", target: 0xC0AEe478e3658e2610c5F7A4A2E1777cE9e4f2Ac, isFactory: true, fee: 30})
        );
        _testCases.push(
            TestCase({network: "ethereum", target: 0x115934131916C8b277DD010Ee02de363c09d037c, isFactory: true, fee: 30})
        );
        _testCases.push(
            TestCase({network: "ethereum", target: 0x9DEB29c9a4c7A88a3C0257393b7f3335338D9A9D, isFactory: true, fee: 30})
        );
        _testCases.push(
            TestCase({network: "ethereum", target: 0x91fAe1bc94A9793708fbc66aDcb59087C46dEe10, isFactory: true, fee: 10})
        );
        _testCases.push(
            TestCase({network: "moonriver", target: 0x28Eaa01DC747C4e9D37c5ca473E7d167E90F8d38, isFactory: true, fee: 30})
        );
    }

    function _setUpRPC(TestCase memory aCase) private {
        // todo: we should also snapshot before each case
        Fork memory lFork = _forks[aCase.network];

        if (lFork.created == false) {
            uint256 lForkId = vm.createFork(vm.rpcUrl(aCase.network));

            lFork = Fork({created: true, forkId: lForkId});

            _forks[aCase.network] = lFork;
            console2.log("fork created for network:", aCase.network);
        }

        vm.selectFork(lFork.forkId);
    }

    function _runCase(TestCase memory aCase) private {
        if (aCase.isFactory == true) {
            (bool lSuccess, uint256 lFee) = _findFee.findFee(IUniswapV2Factory(aCase.target));

            assertEq(lSuccess, aCase.isFactory);
            assertEq(lFee, aCase.fee);
        } else {
            vm.expectRevert();
            _findFee.findFee(IUniswapV2Factory(aCase.target));
        }
    }

    function testFindFee_AllNetworks() external {
        for (uint256 i = 0; i < _testCases.length; ++i) {
            TestCase memory lCase = _testCases[i];

            _setUpRPC(lCase);

            uint256 lBefore = vm.snapshot();
            _findFee = new FindFee();

            _runCase(lCase);

            vm.revertTo(lBefore);
        }
    }
}
````

src/interfaces/IConstantProductPair.sol
xenide: since we're changing the interface in the very near future, this has become irrelevant, resolving for now.

test/unit/AssetManagedPair.t.sol
xenide: alright very nice. I implemented a POC that combines the `snapshot` and `revertTo` approach with the modifier approach you shared previously. 

I can't do this for all the tests yet due to the interface problem. `IAssetManagedPair` doesn't have `mint` and `burn` so I can't convert all of them yet. Once we get the interfaces straight I'll convert all applicable tests into this form

src/curve/stable/StablePair.sol
xenide: Yes I agree 100%. Apologies, was a lil overwhelmed with all the moving parts. 

Enhanced one of the test cases to catch a wrong/missing implementation.

src/interfaces/IAssetManager.sol
OliverNChalk: I would probably want to preserve an interface that the pairs use and AMs **must** implement. This is different to what the owner needs which the AMs **may** implement.

Not sure how to split the different requirements out, have a good name for the two parts of the interface? Maybe:

- `IAssetManager`
- `IAssetManagerExtended`

... not the cleverest

test/unit/AssetManagedPair.t.sol
OliverNChalk: Sounds good, will you delay this merge or merge this and then do a subsequent PR? If doing the subsequent PR, just make sure its tracked either by TODO comment, issue, or personal notes

xenide: yes I will do it in the same PR as the class + interface refactoring. Captured in #76

src/interfaces/IAssetManager.sol
OliverNChalk: Agreed to remove the adjustManagement function entirely. Bots/owner can hit the concrete implementation

xenide: okay removed

OliverNChalk (approved):

PR #77, opened 2022-09-12
refactor: interfaces & classes

### LINKED ISSUE #76: test(pair asset management): make all applicable tests run with modifier `parameterizedTest`
Once the classes and interfaces allow for this

xenide:
## Solution

- note: modified the Burn, Swap event signatures,
   - For Burn, removed the `to`, as there are ERC20 events emitted for the 2 tokens which will indicate what address the tokens have be transferred to 
- with this `swap` interface, it is not possible to flash swap both tokens at the same time, only one token


src/Pair.sol
```diff
@@ -0,0 +1,115 @@
...
+import { GenericFactory } from "src/GenericFactory.sol";
+
+abstract contract Pair is IPair {
+    using FactoryStoreLib for GenericFactory;
+    using Bytes32Lib for bytes32;
+
+    bytes4 private constant SELECTOR = bytes4(keccak256("transfer(address,uint256)"));
+    uint public constant MINIMUM_LIQUIDITY = 10**3;
+
+    uint256 public constant FEE_ACCURACY     = 10_000;
```
OliverNChalk: align

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -248,72 +167,110 @@ contract ConstantProductPair is IConstantProductPair, UniswapV2ERC20, AssetManag
...
+                tokenOut = token1;
+                amountIn = _totalToken0() - _reserve0;
+                require(amountIn == uint256(amount), "CP: AMOUNT_MISMATCH");
+                amountOut = _getAmountOut(amountIn, _reserve0, _reserve1, swapFee);
+            }
+            // swap token1 exact in for token0 variable out
+            else {
+                tokenOut = token0;
+                amountIn = _totalToken1() - _reserve1;
+                require(amountIn == uint256(-amount), "CP: AMOUNT_MISMATCH");
```
OliverNChalk: Does this handle min int256 overflow during cast?

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -248,72 +167,110 @@ contract ConstantProductPair is IConstantProductPair, UniswapV2ERC20, AssetManag
...
+                _safeTransfer(tokenOut == token0 ? token1 : token0, to, actualAmountIn - amountIn);
+            }
+            // do nothing if they are equal
         }
 
-        _update(balance0, balance1, _reserve0, _reserve1);
-        emit Swap(msg.sender, amount0In, amount1In, amount0Out, amount1Out, to);
+        require(_totalToken0() * _totalToken1() >= uint256(reserve0) * reserve1, "CP: K");
+        _safeTransfer(tokenOut, to, amountOut);
+        _update(_totalToken0(), _totalToken1(), _reserve0, _reserve1);
```
OliverNChalk: You need to send the tokens they are owed before you check K.

The flow is:

1. send token
2. callback
3. check K

Realize flashloans are not implemented yet but would've thought the flow would still be the same

src/curve/stable/StablePair.sol
```diff
@@ -236,74 +169,100 @@ contract StablePair is UniswapV2ERC20, ReentrancyGuard, AssetManagedPair {
...
+            // swap token0 variable in for token1 exact out
+            else {
+                tokenOut = token1;
+                amountOut = uint256(-amount);
+                amountIn = _getAmountIn(amountOut, _reserve0, _reserve1, false);
+                actualAmountIn = balance0 - _reserve0;
+            }
+
+            require(amountIn <= actualAmountIn, "SP: INSUFFICIENT_AMOUNT_IN");
+            if (amountIn < actualAmountIn) {
```
OliverNChalk: Unsure how necessary these refunds are, guess it could potentially save the router a CALL to read the reserves?

src/libraries/StableMath.sol
```diff
@@ -51,6 +51,36 @@ library StableMath {
     }
     }
 
+    function _getAmountIn(
```
OliverNChalk, line 54: Where did this impl come from?

xenide: from uniswap / vexchange v2's library. Do we need to attribute it or something?

src/curve/constant-product/ConstantProductPair.sol
xenide: yes this is a mistake... I must have been delirious. Fixed.

src/Pair.sol
xenide: fixed

src/curve/constant-product/ConstantProductPair.sol
xenide: so I just ran some tests of my own, and I realized that at the limit (`type(int256).min`), the cast would return correct results to `uint256`.

I also realized that the minus sign during casting has no effect on the output. Therefore I'm removing `-` from the code

xenide: so I just ran some tests of my own, and I realized that at the limit (`type(int256).min`), the cast would return correct results to `uint256`.

An interesting behavior I found was that the minus sign is not necessary when `amount` is `int256.min`, but is absolutely necessary for all other values.

src/curve/stable/StablePair.sol
xenide: if we don't refund, we just let it sit there and take it as a bonus for the pool?

src/libraries/StableMath.sol
OliverNChalk: This is a stable math impl so cant be uni-v2, was more concerned with its correctness?

src/curve/stable/StablePair.sol
OliverNChalk: yeah, that is what uni-v2 currently does

src/libraries/StableMath.sol
xenide: so this is adapted from:

1. _getAmountOut in the same library
2. Referenced Balancer's implementation [here](https://github.com/balancer-labs/balancer-v2-monorepo/blob/1f07236cd2d3a01fe687c1703b989c6284be2202/pkg/pool-stable/contracts/StableMath.sol#L164) 

But definitely I need to write tests to ensure its correctness

src/curve/stable/StablePair.sol
xenide: okay fixed. We no longer refunding people for giving too much tokens

xenide:
1. implemented the flash swap in both curves. Few caveats:
   - we don't need to do a uni v2 style invariant check now do we? Since all the amounts in and out are controlled by internal functions. 
      - Team agrees, no need for that `K` check since amounts are controlled by `_getAmountIn` and `_getAmountOut`
   - in the current impl, the flash swapper can only pay for the swap with the other 
      - this impl is good for now. 
   - callee interface 
      - keeping the amounts in the interface is good, instead of only `data`

2. corrected the `StableMath::_getAmountIn` implementation
   - the small discrepancy (see tests) between `_getAmountIn` and `_getAmountOut` could be something natural and not something to worry about

src/libraries/StableMath.sol
xenide: implementation corrected with tests written, can review

test/unit/AssetManagedPair.t.sol
```diff
@@ -103,47 +103,81 @@ contract AssetManagedPairTest is BaseTest
...
+        _constantProductPair.burn(address(this));
+
+        // assert
+        assertEq(_manager.getBalance(_constantProductPair, lToken0), 19e18);
+        assertEq(_manager.getBalance(_constantProductPair, lToken1), 19e18);
+        assertLt(_tokenA.balanceOf(address(this)), 10e18);
+        assertLt(_tokenB.balanceOf(address(this)), 10e18);
+    }
+
+    function testSyncManaged_Stable() external
```
OliverNChalk, line 180: Do these two tests still need to be separate or can be paramterized?

OliverNChalk (approved):

test/unit/AssetManagedPair.t.sol
xenide: in this case they need to be separate, mainly due to: 

1. the liquidity returned on L.158 and L.192 are different, because the way liquidity is calculated is diff for constant product pair vs stable pair (sqrt vs addition)
   - I have thought of using a check to see if the current `_pair` is `CP` or `SP`, but can't find ways to do it easily/reliably so decided to separate the tests for now. 
2. because the liquidity means different things for `SP` and `CP`, when we burn the same amount of LP tokens, the amount of underlying returned would be greatly different too

xenide:
Merging despite the integration tests failing. This is due to AAVE suspending deposits on fantom in preparation for pulling out of the network. See this [gov proposal](https://governance.aave.com/t/arc-aave-v3-fantom-freeze-reserves/9166) for details.

PR #78, opened 2022-09-13
fix: oracle price reverts when difference > `1e18`

### LINKED ISSUE #69: fix(OracleMath): will fail if price between the two tokens exceed 1e18
Discovered during fuzz testing. Might not be fixable 

As the `LogCompression::toLowResLog` library takes in a value that is scaled by 1e18 (therefore as 1 Wei as precision), if the prices of the tokens in the pair go so far away that it goes beyond the 1e18 scale, all swaps / burns / mints will fail

Another scenario to note: when the decimals of two pairs are not the same (in ConstantProductPair), this problem might be exacerbated, e.g. between USDC (6dp) and a token with 18dp

xenide:
## Solution

- use `divWadUp` to ensure that the result is at least 1


OliverNChalk:
Can we rename pr `fix: oracle price now does not fail when price difference is greater than 1e18` -> `fix: oracle price reverts when difference > 1e18`, as it stands it sounds like you're making it revert now when the opposite is true

src/libraries/ConstantProductOracleMath.sol
```diff
@@ -14,20 +14,27 @@
 
 pragma solidity 0.8.13;
 
+import "solmate/utils/FixedPointMathLib.sol";
```
OliverNChalk: Use explicit import

src/libraries/ConstantProductOracleMath.sol
```diff
@@ -14,20 +14,27 @@
...
     function calcLogPrice(
         uint256 reserve0,
         uint256 reserve1
     ) internal pure returns (int112 logSpotPrice) {
         // scaled by 1e18
-        // reserve0 and reserve1 should never be 0, as checked by _update()
-        uint256 spotPrice = reserve1 * 1e18 / reserve0;
+        // spotPrice will never be zero as we do a divWadUp
+        // the minimum price would be 1 wei (1e-18)
+        uint256 spotPrice = reserve1.divWadUp(reserve0);
```
OliverNChalk, line 38: This is what balancer does right? That means the min price is `1e-18 A of B` and vice versa. This seems totally fine but we should document it somewhere?

src/libraries/ConstantProductOracleMath.sol
```diff
@@ -14,20 +14,27 @@
...
 import "src/libraries/Math.sol";
 import "src/libraries/LogCompression.sol";
 
 library ConstantProductOracleMath {
+    using FixedPointMathLib for uint256;
+
     /**
      * @dev Calculates the spot price of token1/token0 for the constant product pair
+     * @param reserve0 should never be 0, as checked by _update()
+     * @param reserve1 should never be 0, as checked by _update()
```
OliverNChalk:
```suggestion
     * @notice Calculates the spot price of token1/token0 for the constant product pair
     * @dev Minimum price is 1e-18, as we do not round to zero
     * @param reserve0 should never be 0, as checked by _update()
     * @param reserve1 should never be 0, as checked by _update()
```

src/libraries/ConstantProductOracleMath.sol
xenide: fixed all imports in this file as well as in its parallel `StableOracleMath.sol`

src/libraries/ConstantProductOracleMath.sol
xenide: okay given that you have suggested the change in the function docs, I think that should be sufficient.

xenide:
> Can we rename pr `fix: oracle price now does not fail when price difference is greater than 1e18` -> `fix: oracle price reverts when difference > 1e18`, as it stands it sounds like you're making it revert now when the opposite is true

okay changed the title

OliverNChalk (approved):

PR #82, opened 2022-09-19
feat: oracle for `StablePair`

xenide:
## Solution

- implemented the oracle logic for StablePair 
- changed the return structures of functions in `StableOracleMath` a little to optimize the calculations (now we avoid calculating the invariant twice)


src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -259,7 +259,7 @@ contract ConstantProductPair is ReservoirPair {
         unchecked {
             int112 logAccPrice = previous.logAccPrice + currLogPrice * int112(int256(uint256(timeElapsed)));
             int112 logAccLiq = previous.logAccLiquidity + currLogLiq * int112(int256(uint256(timeElapsed)));
-            index += 1;
+            ++index;
```
OliverNChalk: The former is more readable?

src/curve/stable/StablePair.sol
```diff
@@ -404,7 +419,16 @@ contract StablePair is ReservoirPair {
...
-        // todo: implement this
+    function _updateOracle(uint256 _reserve0, uint256 _reserve1, uint32 timeElapsed, uint32 timestampLast) internal override {
+        Observation storage previous = observations[index];
+
+        (int112 currLogPrice, int112 currLogLiq) = StableOracleMath.calcLogPriceAndLiq(_getCurrentAPrecise(), _reserve0, _reserve1);
+
+        unchecked {
+            int112 logAccPrice = previous.logAccPrice + currLogPrice * int112(int256(uint256(timeElapsed)));
+            int112 logAccLiq = previous.logAccLiquidity + currLogLiq * int112(int256(uint256(timeElapsed)));
+            ++index;
```
OliverNChalk: index += 1 if we change the above code too

src/libraries/StableOracleMath.sol
```diff
@@ -6,32 +6,36 @@ import { Math } from "src/libraries/Math.sol";
...
+    ) internal pure returns (int112 logSpotPrice, int112 logLiq) {
+        (uint256 spotPrice, uint256 liq) = calcSpotPrice(amplificationParameter, reserve0, reserve1);
+
+        int256 rawLogSpotPrice = LogCompression.toLowResLog(spotPrice);
+        assert(rawLogSpotPrice >= type(int112).min && rawLogSpotPrice <= type(int112).max);
+        logSpotPrice = int112(rawLogSpotPrice);
+
+        int256 rawLogLiq = LogCompression.toLowResLog(liq);
+        assert(rawLogLiq >= type(int112).min && rawLogLiq <= type(int112).max);
+        logLiq = int112(rawLogLiq);
```
OliverNChalk, line 28: Does liquidity grow if we increase amplification & is this what we want if it does?

src/oracle/OracleWriter.sol
```diff
@@ -6,5 +6,11 @@ abstract contract OracleWriter is IOracleWriter {
...
     uint16 public index = type(uint16).max;
 
-    function _updateOracle(uint112 _reserve0, uint112 _reserve1, uint32 timeElapsed, uint32 timestampLast) internal virtual;
+    /**
+     * @param _reserve0 already normalized to 18 decimals
+     * @param _reserve1 already normalized to 18 decimals
+     * @param timeElapsed time since the last oracle observation
+     * @param timestampLast the time of the last activity on the pair
+     */
+    function _updateOracle(uint256 _reserve0, uint256 _reserve1, uint32 timeElapsed, uint32 timestampLast) internal virtual;
```
OliverNChalk, line 15: Do we need to pass timeElapsed or is it just a simple `block.timestamp - timestampLast`?

src/curve/constant-product/ConstantProductPair.sol
xenide: okay fixed. 

I even ran a test to see if either is more gas efficient. They're equivalent.

src/curve/stable/StablePair.sol
xenide: fixed

src/libraries/StableOracleMath.sol
xenide: So the way the invariant is calculated is: 

- it has a max of `reserve0` + `reserve1`. This happens when the they are perfectly balanced (i.e. `reserve0` == `reserve1`)
- when the amounts are imbalanced, the greater the A, the closer it is to the max of `reserve0` + `reserve1`
- For most of the time, when the two amounts are close, the invariant doesn't even vary that much according to the amp coefficient

Therefore, I think it is okay to use this invariant as the measure for liquidity. 

The other alternative is, we just straight up add the two token amounts together. But this is not recommended as, when the ratio becomes really skewed (such as when a stablecoin depegs), the actual value of the liquidity is no longer just `reserve0` + `reserve1`, but much lower. Our current invariant calculation can provide for a lower liquidity when a shit stablecoin depegs.

src/oracle/OracleWriter.sol
xenide: the passed in `timeElapsed` allows for overflow, and is built upon the `blockTimestamp`, which is casted to `uint32` and modded `2 ** 32`. 

if we only pass in `timestampLast`, we need to do the casting to `uint32` and `% 2**32` again on `block.timestamp`

src/libraries/StableOracleMath.sol
OliverNChalk: Okay sounds good, happy to resolve and move on. We can always finalize in audit/later testing

src/oracle/OracleWriter.sol
OliverNChalk: Okay, happy to pass in - makes sense

OliverNChalk (approved):

PR #83, opened 2022-09-23
fix: oracle decimal diff

### LINKED ISSUE #79: fix(constant product oracle): support calculating prices of tokens which do not have 18 decimals
Currently, if the tokens do not have 18 decimals, the oracle does not take that into account. We need to normalize it to 18 decimals before calculation to yield an accurate price

xenide:
## Motivation

- previously, if the decimals of the tokens differed, the oracle calculations would not take them into account

## Solution

- now, the oracle calculation takes into account the decimal places by normalizing them all to 18 decimal places
- however, a side effect of this is that, tokens with more than 18 decimals will not be supported, as demonstrated by the test case


src/curve/stable/StablePair.sol
```diff
@@ -268,8 +268,8 @@ contract StablePair is ReservoirPair {
 
         if (timeElapsed > 0 && _reserve0 != 0 && _reserve1 != 0) {
             _updateOracle(
-                uint256(_reserve0) * token0PrecisionMultiplier,
-                uint256(_reserve1) * token1PrecisionMultiplier,
+                uint256(_reserve0),
+                uint256(_reserve1),
                 timeElapsed,
```
OliverNChalk, line 273: What was the reason for the change, to align across pool types?

xenide: yes it's to standardize across pool types. See the doc changes for `_updateOracle`.

OliverNChalk (approved):

PR #84, opened 2022-09-27
fix: wrong token out

xenide:
## Motivation

- discovered when writing v3-periphery tests


OliverNChalk:
Can you add a v3-core test that flags this issue before fixing it?

xenide:
> Can you add a v3-core test that flags this issue before fixing it?

I have added the test which validates the correctness of the current implementation, and in the absence of the fix, will fail. Please review. 

For `StablePair`, there is already a test case for this scenario so there is no need to add additional ones.

test/unit/ConstantProductPair.t.sol
```diff
@@ -267,6 +279,46 @@ contract ConstantProductPairTest is BaseTest
...
+    {
+        // arrange
+        vm.prank(address(_factory));
+        _constantProductPair.setCustomSwapFee(0);
+        uint256 lMinNewReservesOut = INITIAL_MINT_AMOUNT ** 2 / type(uint112).max + 1;
+        // this amount makes the new reserve of the input token stay within uint112 and not overflow
+        uint256 lMaxOutputAmt = INITIAL_MINT_AMOUNT - lMinNewReservesOut;
+        uint256 lAmountOut = bound(aAmountOut, 1, lMaxOutputAmt);
+        (uint256 lReserve0, uint256 lReserve1, ) = _constantProductPair.getReserves();
+        uint256 lAmountIn = _calculateInput(lReserve0, lReserve1, lAmountOut, _constantProductPair.swapFee());
```
OliverNChalk:
```suggestion
        // assume
        uint256 lMinNewReservesOut = INITIAL_MINT_AMOUNT ** 2 / type(uint112).max + 1;
        // this amount makes the new reserve of the input token stay within uint112 and not overflow
        uint256 lMaxOutputAmt = INITIAL_MINT_AMOUNT - lMinNewReservesOut;
        uint256 lAmountOut = bound(aAmountOut, 1, lMaxOutputAmt);
        
        // arrange
        vm.prank(address(_factory));
        _constantProductPair.setCustomSwapFee(0);
        (uint256 lReserve0, uint256 lReserve1, ) = _constantProductPair.getReserves();
        uint256 lAmountIn = _calculateInput(lReserve0, lReserve1, lAmountOut, _constantProductPair.swapFee());
```

OliverNChalk: Make sure `vm.assume` or `bound` is always at top of test or even in its own section

OliverNChalk (approved):
Approved after minor re-ordering

test/unit/ConstantProductPair.t.sol
xenide: noted, will do so in the future. Fixed this as well as the corresponding test for `StablePair`. 

Will also do a refactor in the future to take care of all the cases where we use `bound` or `assume`.

PR #86, opened 2022-10-04
fix: `StablePair::_mintFee`

### LINKED ISSUE #44: test: mintFee calc during ramping
`_mintFee` calculates the growth of invariant D in between liquidity addition and removal. However, the value of D is dependent on A when the pool is imbalanced. We need more testing of this, to ensure that the amplification coefficient A does not affect the platformFee calculation

## Deliverables 
- [ ] PR with additional tests

xenide:
## Motivation

- turns out that my previous approach of storing the `lastLiquidityEventReserves` is correct, but is hard to prove. Thus the change in the approach

- in the test case with the prev impl, when the pools are balanced, platform fee received difference is small
   - diff gets big when the pools go out of balance, we see a 17% diff in platform fee collected when the ratio of the tokens become 3:1 

## Solution

- adopting balancer's approach, which is to store `lastInvariant` and `lastInvariantAmp`, instead of `lastLiquidityEventReserves`
- I have also done manual calculations to make sure that the amount minted to `platformFeeTo` is in accordance with the `platformFee` percentage (~ 25% in the tests )


src/curve/stable/StablePair.sol
```diff
@@ -150,7 +152,7 @@ contract StablePair is ReservoirPair {
     function burn(address to) public nonReentrant returns (uint256 amount0, uint256 amount1) {
         _syncManaged();
 
-        (uint256 balance0, uint256 balance1) = _balance();
+        (uint256 _reserve0, uint256 _reserve1, ) = _getReserves();
```
OliverNChalk: Unrelated but StablePair has _getReserves() and getReserves(). Should we delete the private one and just use the public one?

src/curve/stable/StablePair.sol
```diff
@@ -159,24 +161,24 @@ contract StablePair is ReservoirPair {
...
         _burn(address(this), liquidity);
         _safeTransfer(token0, to, amount0);
         _safeTransfer(token1, to, amount1);
 
         _update(_totalToken0(), _totalToken1(), reserve0, reserve1);
 
-        lastLiquidityEventReserve0 = reserve0;
-        lastLiquidityEventReserve1 = reserve1;
+        lastInvariant = uint192(_computeLiquidity(reserve0, reserve1));
+        lastInvariantAmp = _getCurrentAPrecise();
```
OliverNChalk, line 204: Can gas be saved using `getReserves` to batch fetch - or does solidity already optimize this under the hood?

OliverNChalk (approved):

src/curve/stable/StablePair.sol
xenide: oh wow I did a test by replacing all instances of the private function with the public one. There is ZERO gas difference. I guess I'll delete the private one and just use the public one. 

fixed

src/curve/stable/StablePair.sol
xenide: I tried doing 

```
(_reserve0, _reserve1, ) = getReserves();
lastInvariant = uint192(_computeLiquidity(_reserve0, _reserve1));
```

somehow it actually costs (a little) more gas. Note that the variables are not new, they're re-used. When I use new variables the results are the same

xenide: therefore, resolving for now

PR #87, opened 2022-10-05
fix: non-optimal mint ratio

xenide:
## Motivation

- during the initial phases of importing the code from sushi-trident and refactoring, I deleted the `_nonOptimalMintFee` function as I thought it served no use. However, I realized afterwards that it is crucial to ensuring that we charge fees for adding of liquidity out of the expected proportion. Thus we're bringing back this function

## Solution

- bring back the function with a `require` on the limits of the return value to prevent overflow 
- test cases demonstrate that the user get charged fees when adding in non optimal proportions


OliverNChalk:
Hmm, I'm not a fan of imbalanced liquidity provision because:

1. There is an implicit swap (that users may not be aware of)
2. It can be achieved via router multicall or helper contract
3. It bloats the core code for what is essentially a helper function

Let's discuss the pros and cons in person

OliverNChalk (approved):

PR #89, opened 2022-10-08
fix: integration test

### LINKED ISSUE #85: fix: integration tests
the integration tests are failing because fantom is removing support for AAVE. The fix will likely integrate with AAVE on another chain (such as AVAX)

xenide:
## Motivation

- previous integration test broke because AAVE is removing support for FTM 
- while refactoring, realized that I did not handle the case whereby there is no market for the token. So I have added a check for that. 
- closes #85 

## Solution

- in tests, moved statements with a `bound` under an `// assume` header
- in tests, cater for token0/1 when calling functions
- checking for money market for individual tokens and acting accordingly 
- forking 2 networks and testing for now (as a POC). Once we confirm the networks we want to deploy on, we will add them to the integration test

## Others

- AaveManager::L15, is this necessary? Cuz it's just an event, not a function
- fun fact: when fuzzing, somehow if you pass in an address, it is significantly slower than passing in an `uint256`

### DIFF
--- a/test/integration/Aave.t.sol
+++ b/test/integration/Aave.t.sol
@@ -101,7 +101,7 @@ contract AaveIntegrationTest is BaseTest
         _pairs.push(_stablePair);
     }
 
-    function setUp() public
+    function setUp() external
     {
         _networks.push(
             Network(vm.rpcUrl("avalanche"), 0xB97EF9Ef8734C71904D8002F8b6Bc66Dd9c48a6E)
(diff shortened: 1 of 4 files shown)


test/integration/Aave.t.sol
```diff
@@ -55,161 +101,236 @@ contract AaveIntegrationTest is BaseTest
         _pairs.push(_stablePair);
     }
 
+    function setUp() public
```
OliverNChalk:
```suggestion
    function setUp() external
```

Favour external when public is not necessary (i.e. func is not called within contract).

OliverNChalk (approved):

test/integration/Aave.t.sol
xenide: okay so we should ensure this even for tests as well. Fixed and noted.

PR #90, opened 2022-10-09
feat: factory allPairs array 

xenide:
## Motivation

- there is a need for the factory to keep track of all the pairs that have been created, for the frontend and other data consumers


src/GenericFactory.sol
```diff
@@ -42,6 +42,12 @@ contract GenericFactory is IGenericFactory, Ownable
     event PairCreated(address indexed token0, address indexed token1, uint256 curveId, address pair);
 
     mapping(address => mapping(address => mapping(uint256 => address))) public getPair;
+    address[] public allPairs;
+
+    function allPairsLength() external view returns (uint256)
+    {
+        return allPairs.length;
+    }
```
OliverNChalk: Thoughts on returning the array? `allPairs returns (memory address[])`

xenide: yes makes sense. Cuz the consumers would likely want the whole array rather than accessing it at a specific index. 

Fixed it

OliverNChalk (approved):

PR #92, opened 2022-10-14
test: more StablePair test cases

### LINKED ISSUE #45: test: furnish test cases for Stableswap with inspirations from other projects
Currently, we do not have any legacy tests for stableswap. All tests have been written from scratch. However, to gain more confidence in the correctness of the code and our changes, we should test it as best as we can. To that end, we should refer to existing project with stableswap, adopt relevant tests, and re-write them in foundry. 

Some projects to refer to when sourcing for test cases
- Sushi trident's original tests
   - we were running these, but once we threw away the dependency on bentobox etc it no longer works 
   - reviewed them again recently in mid Oct 2022. Seems that their cases are pretty minimal, nothing we don't already have 
- Curve finance
- Saddle finance
- Balancer

xenide:
## Motivation

- as the StablePair doesn't have a legacy suite of test cases unlike the ConstantProductPair, it is important that we build a robust suite of test cases to ensure the correct implementation and gain confidence when optimizing for performance in the near future

## Solution

- added many fuzzed and non-fuzzed test cases to cover all grounds, as inspired by the other stableswap projects

## Notes

- while writing test cases, I noticed that currently `ConstantProductPair` does not allow a zero burn, while `StablePair` allows a zero burn. Do you think we need to change this? 
  - an argument for not allowing zero burns is, it prevents mistakes as some EOA / smart contract might forget to transfer the LP tokens into the pair before calling `burn()`. Another argument for that is, since we don't allow zero mints on both pairs, why should we allow zero burns? 

- noticed that we might have to change our `burn` implementation again. Although we do a try catch in the `mintFee`, we're still doing a `_computeLiquidity` at the end of the function, which has a possibility of reverting
  - so we have several options: 
    1. make the iterative function `StableMath._computeLiquidityFromAdjustedBalances` not revert, and return whatever value at the end of the 256 iterations
       1. in my fuzz tests, it seems that the diff between the last and second last iterated value is a few Wei, maximum of around 20
       2. sushi-trident does this, but their code is neither audited nor in production
    3. go back to my previous implementation of storing `lastLiquidityEventReserve0/1`. That way, we avoid another call to the iterative function and `burn` would never revert  
  - second option is recommended 
  - actually, I also realized that even if we fix this, `burn` would still attempt to call the iterative function again in `_update`

- after writing the test case `testAttackWhileRampingDown`, we should take note of a few things in operations
  - when ramping A (up or down), make sure that the swap fee is not 0. If not, it makes the pool more exploitable
  - especially when ramping down, we need to give it enough time. This is already taken care of by the rate limit of A change in the code


src/curve/stable/StablePair.sol
```diff
@@ -45,15 +45,13 @@ contract StablePair is ReservoirPair {
     {
         ampData.initialA        = factory.read("ConstantProductPair::amplificationCoefficient").toUint64() * uint64(StableMath.A_PRECISION);
         ampData.futureA         = ampData.initialA;
-        // perf: check if intermediate variable is cheaper than two casts (optimizer might already catch it)
         ampData.initialATime    = uint64(block.timestamp);
         ampData.futureATime     = uint64(block.timestamp);
```
OliverNChalk, line 50: You try ampData.futureATmie = ampData.initialATime?

src/GenericFactory.sol
```diff
@@ -62,7 +62,6 @@ contract GenericFactory is IGenericFactory, Ownable
         require(aTokenA != address(0), "FACTORY: ZERO_ADDRESS");
         require(getPair[aTokenA][aTokenB][aCurveId] == address(0), "FACTORY: PAIR_EXISTS");
         address lCodePointer = _getByteCode[aCurveId];
-        require(lCodePointer != address(0), "FACTORY: INVALID_CURVE_ID");
```
xenide, line 65: removed this line because if `aCurveId` was invalid, the caller would get an index out of bound error and would not reach the `require` statement. Also added a corresponding test case for that. 

The `require` would make sense if `_getByteCode` were a mapping but here it's an array

package.json
```diff
@@ -65,7 +63,6 @@
     "test": "npm run test:unit",
     "test:all": "npm run test:unit && npm run test:integration",
     "test:integration": "FOUNDRY_PROFILE=integration forge test ",
-    "test:trident": "npm run build:trident && ts-node scripts/inject_trident.ts && cd reference/sushi-trident && yarn run test",
```
OliverNChalk, line 68: We can delete the inject_trident script?

test/unit/libraries/StableMath.t.sol
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    function testComputeLiquidityFromAdjustedBalances_ConvergeEvenWithVeryUnbalancedValues (
+        uint256 aReserve0, uint256 aReserve1, uint256 aN_A
+    ) public
+    {
+        // assume - covers ratios up to 1:1000, which is good enough even in the case of a depeg
+        uint256 lReserve0 = bound(aReserve0, 1e6, type(uint112).max);
+        uint256 lReserve1 = bound(aReserve1, lReserve0 / 1e3, Math.min(type(uint112).max, lReserve0 * 1e3));
+        uint256 lN_A = bound(aN_A, StableMath.MIN_A * StableMath.A_PRECISION, StableMath.MAX_A * StableMath.A_PRECISION);
```
OliverNChalk: What happens if it goes out of range, it just reverts?

test/unit/libraries/StableMath.t.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint256 lN_A = 2 * bound(aAmp, StableMath.MIN_A * StableMath.A_PRECISION, StableMath.MAX_A * StableMath.A_PRECISION);
+
+        // arrange
+        uint256 lReserve0 = 120_000_000e18;
+        uint256 lReserve1 = 11_000_000e6;
+
+        // act
+        uint256 lAmtOut = StableMath._getAmountOut(lAmtIn, lReserve0, lReserve1, 1, 1e12, false, lSwapFee, lN_A);
+
+        // assert - what to assert?
```
OliverNChalk: in the test below you assert that the price for execution cannot go above 1. Do you wanna do something similar here? Set a high price and then assert the price cannot go below 1

src/curve/stable/StablePair.sol
xenide: yup I tried that and storing into an intermediate variable. Did not become more gas efficient, that's why I removed the comments

package.json
xenide: okay deleted

test/unit/libraries/StableMath.t.sol
xenide: okay so here's the thing, having fuzzed a lot of iterative functions, I noticed that the function does not converge on some random values sometimes. On average once every 150 fuzzes

It is not that when the difference between two values goes beyond a certain threshold then it fails. Sometimes relatively close values (within `1e6`) can also fail. On the other hand differences beyond `1e22` can still converge. 

Examples of values that don't converge 

```
4280926068553959580125235477587551
428092606855395958012523547758

17515493530788801010825324439190
1751549353078880101082532454

1001927
103

974523083462972163360456
97452308346297216570

1000000
13

2828486865529418933375394711587169
116901276727300513218678699616

1000000
4874083481260427647068269489

2244282415425521217040721616
1000000
```

I don't know what's this pattern... Every time the ratios are close to some power of 10 it fails? Or when one number is 1e21 larger than the other number? 

In my study of the other projects, 

- balancer literally has **ONE** test case where they take a very large number, a very small number and show that it converges. No fuzzing
- Saddle doesn't even have a test case to show that any numbers will converge.  
- sushi trident didn't even test any of these stable math functions 
- curve also didn't have a single test case for this invariant function

So we're literally fuzzing more than all of them put together. Incredible right?

test/unit/libraries/StableMath.t.sol
xenide: still working on this. But given the correct conditions (sufficiently high fees and high enough amplification coefficient), the amount out can actually be slightly less than the amount in, despite the ratio being skewed towards one token. 

I don't think there's anything wrong with the math. Just the amalgamation of conditions.

test/unit/libraries/StableMath.t.sol
OliverNChalk: It enters a loop before 100 iterations and stays there forever. Seems cyclical:

image

It appears to have two causes:

1. Sometimes can only get within 2 (could just be an unlucky rounding scenario).
2. Has a maximum accuracy, for instances HurHur2 is accurate within 0.0000000000000029%.

I see a few options:

1. Keep the revert if we assume it exceedingly rare for that ratio to reach this level of imbalance.
  a. What remedies do we have if someone manipulates it into this imbalanced state? Is it possible to get funds locked forever?
2. Instead of within1, do withinAccuracy
  b. Or you could still do within1 but at the end instead of reverting, if its within tolerance you return it.
3. Raise an issue with a failing test case on balancer repo, let their maths gigabrains solve it for us

xenide: This is a difficult question, let us take some time to think through before setting on a solution. Because this decision is critical to the design. 

If somehow the pool gets into this state, we can get it out of that state by, the `burn` function, which is why it is important that it has a failsafe to ensure that it always succeeds. It lodges the state out of that death loop. But this is just my conjecture. I don't have a test to prove that this is indeed the case. 

In my searching of curve, saddle, balancer's git repos and discords, no one seems to have hit this situation in production. But in beethovenx on fantom (balancer fork), they have [hit this problem](https://discord.com/channels/885764705526882335/886037144211173386/973957616755691590) before when UST depegged.

So we see that this will happen when a stablecoin depegs, which is a matter of when not if for us. 

That said, I think option 2 is a good middle ground. The inaccuracy is usually so small that it does not affect the calculations in a very material way. We can define a small enough window such ensure that. The return value can be `(y + yPrev / 2)`. If the pool gets into a state that it doesn't fit into this window, then it is truly fucked

OliverNChalk: Okay, let's discuss in person then

xenide:
For the convergence problem, I have implemented the threshold solution. Given the following conditions, this threshold is the tightest we can go before it fails: 
- Threshold: `0.0000000000004e18` (equivalent to 0.000000004 basis points) 
- Conditions:
    - `uint256 lReserve0 = bound(aReserve0, 1e18, type(uint112).max / 100)`
    - `uint256 lReserve1 = bound(aReserve1, lReserve0 / 1e10, Math.min(type(uint112).max / 100, lReserve0 * 1e10))` 
- Given the huge ranges we have done for fuzzing (and the number of runs I have done), I am confident that this threshold will serve us well in production


However, it is worthy to note that the threshold is dependent on the conditions as well. When the reserves get really small (< `1e18`), the relative error will increase and it will go beyond our acceptable threshold and revert. However, I think we can forgo those cases, because if our pools were locked with < 1e18 (normalized) tokens forever it’s hardly the end of the world. 

My only remaining concern is, what will the auditors say about this. 

If we solve this, then we essentially don’t need the `try/catch` for the mintFee for `burn` anymore.

--- 

Other interesting things I found out while testing: 

1. Sometimes if the numbers are too big, for example near the max of `uint112`, the calculation actually overflows
- It is during the calculating of the D term, esp the operation `((N_A * s) / A_PRECISION + 2 * dP) * D`
- If I set the fuzz limit to `type(uint112).max / 100` it’s fine, it will definitely not overflow 
- In real life scenarios this is very unlikely to happen to us
- I have also tried plugging those huge numbers into curve calc algo and they overflow as well

src/libraries/StableMath.sol
```diff
@@ -103,6 +104,13 @@ library StableMath {
                 return D;
             }
         }
+        /// @dev sometimes the iteration gets stuck in repeating loop
+        /// so if it is close enough we return it anyway
```
OliverNChalk: Natspec is for interfaces, should use regular comment

OliverNChalk (approved):

src/libraries/StableMath.sol
xenide: okay fixed

PR #95, opened 2022-10-21
feat: diff default swap fees for diff curves

### LINKED ISSUE #91: feat: differentiate default swap fees for ConstantProduct and Stable
Currently the default swap fee applies to all pairs. However, it would be advantageous to introduce default swap fees for different curve types, as stable pairs usually would charge lower fees compared to volatile pairs.

xenide:
## Solution

- only differentiated the swap fee for diff curves. Platform fee remains shared between the curves


src/Pair.sol
```diff
@@ -73,7 +71,7 @@ abstract contract Pair is IPair, UniswapV2ERC20 {
         emit CustomSwapFeeChanged(customSwapFee, _customSwapFee);
         customSwapFee = _customSwapFee;
 
-        updateSwapFee();
+        this.updateSwapFee();
```
OliverNChalk: Why use `CALL` instead of just an abstract function?

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -21,7 +21,23 @@ contract ConstantProductPair is ReservoirPair {
     uint224 public kLast; // reserve0 * reserve1, as of immediately after the most recent liquidity event
 
     constructor(address aToken0, address aToken1) Pair(aToken0, aToken1)
-    {} // solhint-disable-line no-empty-blocks
+    {
+        swapFee = uint256(factory.get(keccak256("CP::swapFee")));
+
+        require(swapFee <= MAX_SWAP_FEE, "CP: INVALID_SWAP_FEE");
```
OliverNChalk: I think you can DRY this with either an abstract constant or abstract function. Like `abstract constant PAIR_SWAP_FEE_NAME: string`

src/curve/stable/StablePair.sol
```diff
@@ -43,12 +43,14 @@ contract StablePair is ReservoirPair {
...
     {
-        ampData.initialA        = factory.read("ConstantProductPair::amplificationCoefficient").toUint64() * uint64(StableMath.A_PRECISION);
+        ampData.initialA        = factory.read("SP::amplificationCoefficient").toUint64() * uint64(StableMath.A_PRECISION);
         ampData.futureA         = ampData.initialA;
         // perf: check if intermediate variable is cheaper than two casts (optimizer might already catch it)
         ampData.initialATime    = uint64(block.timestamp);
         ampData.futureATime     = uint64(block.timestamp);
 
+        swapFee = uint256(factory.get(keccak256("SP::swapFee")));
+
```
OliverNChalk: Can be DRY

OliverNChalk (changes requested):

src/Pair.sol
xenide: yeah so it's either this way or we have to define the abstract function in `Pair.sol` and then we have to do the `override` in `ConstantProductPair / StablePair`

```solidity 
    function updateSwapFee() public override(IPair, Pair) 
```
which I thought was a little ugly, which is the reason I chose using the `this` approach

using `this` doesn't make it an external call does it?

src/curve/constant-product/ConstantProductPair.sol
xenide: okay so we can't explicitly define an `abstract constant` in `Pair.sol` and then use it in the same contract. The closest thing is declaring it in `IPair.sol` and then defining it in ConstantProduct/Stable. 

So we got 2 options
1. define a `PAIR_SWAP_FEE_NAME` in `IPair.sol`. Then we still have to call `this.PAIR_SWAP_FEE_NAME()` in `updateSwapFee`. Cuz it's not defined in `Pair.sol` but in the individual curves. 
2. we define a `string internal swapFeeName` in `Pair.sol`, and then set that at the constructor of each curve. I tried making this immutable. We cannot even if it is `bytes32` because this variable will be set in the individual curves, not the abstract contract itself. Immutable variables can only be set in the constructor of the defining contract. 
   - this is the current impl

src/curve/stable/StablePair.sol
xenide: fixed, also changed all `.get` to the more civilized `.read()`

src/Pair.sol
OliverNChalk: It does make it an external call, debug and take a look if you want (or easier, run with `-vvvv`)

src/curve/constant-product/ConstantProductPair.sol
OliverNChalk: 2 is better, avoid CALL. Solidity sucks for not letting you use abstract consts in the contract that declares them... What about if you define it as a constructor argument for Pair.sol so that the child is forced to declare its curve name to the parent when it constructs the parent?

Then it can be immutable

src/Pair.sol
xenide: okay indeed it was an external call. With a gas cost of `3880`. So we won't do that.

src/curve/constant-product/ConstantProductPair.sol
xenide: Passing the `swapFeeName` as a parameter into the base class is a genius idea. 

Except that for solidity right now it can't have non-value types as immutables. Meaning we can't have `string` immutables, only `bytes32`. So strings can be constants but can't be immutables. 

So anyway, to make this work, we have to store `swapFeeName` as a bytes32, which is:

```solidity
swapFeeName = keccak256(abi.encodePacked(aPairSwapFeeName)); 
```

So this would differ from the other constant types, which are string. And then we have to use `.get()` for this case instead of `.read()`. Both of which I can live with. See latest commit

OliverNChalk: Works for me. I have a feeling during gas golfing we may move away from strings to bytes32 strings or something anyway. So let's leave it for now. Can resolve

src/Pair.sol
```diff
@@ -16,6 +16,9 @@ abstract contract Pair is IPair, UniswapV2ERC20 {
 
     bytes4 private constant SELECTOR        = bytes4(keccak256("transfer(address,uint256)"));
     uint256 public constant MINIMUM_LIQUIDITY  = 10**3;
+    string private constant PLATFORM_FEE_NAME = "Shared::platformFee";
+    string private constant RECOVERER_NAME = "Shared::defaultRecoverer";
+    string internal constant PLATFORM_FEE_TO_NAME = "Shared::platformFeeTo";
```
OliverNChalk: The alignment looks a little drunk here :P

OliverNChalk (approved):

src/Pair.sol
xenide: rearranged and re-aligned them. Should be a lot better now 😃

PR #96, opened 2022-10-25
feat: asset manager `returnAsset` and handling failure conditions

xenide:
## Motivation
- see [issue](https://github.com/vexchange/v3-core/issues/98) 
- in the process of writing tests, I realized that we have not handled certain conditions that might arise, such as the AAVE vault being [frozen, paused](https://discord.com/channels/602826299974877205/636902500041228309/1031515565740806204), or simply reached its [supply cap](https://docs.aave.com/developers/getting-started/readme#supply-and-borrow-caps). So we need to handle these cases too 

## Solution

- in the current solution, the pair does not pull more than it has to
   - this way, more tokens can remain in the yield generating protocol, increasing the APY

- effect on gas seems to be modest 
 
TODO: 

- [x]  skim method also uses `_safeTransfer` so it needs to be refactored as well


src/asset-management/AssetManagedPair.sol
```diff
@@ -127,12 +127,5 @@ abstract contract AssetManagedPair is Pair, IAssetManagedPair {
 
             IERC20(token1).transferFrom(address(assetManager), address(this), lDelta);
         }
-
-        _update(
-            _totalToken0(),
-            _totalToken1(),
-            reserve0,
-            reserve1
-        );
```
xenide, line 136: removed this as there is no need it for it. The reserves would still be the same before and after the transfer to AM. It's only the managed amounts that change

src/asset-management/AaveManager.sol
```diff
@@ -168,6 +168,15 @@ contract AaveManager is IAssetManager, Ownable, ReentrancyGuard
...
 
+    function returnAsset(address aToken, uint256 aAmount) external {
+        require(aAmount > 0, "AM: ZERO_AMOUNT_REQUESTED");
+        IAssetManagedPair lPair = IAssetManagedPair(msg.sender);
+        int256 lAmount0Change = -int256(aToken == lPair.token0() ? aAmount : 0);
+        int256 lAmount1Change = -int256(lAmount0Change == 0 ? aAmount: 0);
+        assert(lAmount0Change <= 0 && lAmount1Change <= 0);
+        _adjustManagement(lPair, lAmount0Change, lAmount1Change);
+    }
+
```
OliverNChalk: Authentication?

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -167,8 +167,28 @@ contract ConstantProductPair is ReservoirPair {
         amount1 = liquidity * balance1 / _totalSupply; // using balances ensures pro-rata distribution
         require(amount0 > 0 && amount1 > 0, "CP: INSUFFICIENT_LIQ_BURNED");
         _burn(address(this), liquidity);
-        _safeTransfer(_token0, to, amount0);
-        _safeTransfer(_token1, to, amount1);
+
+        if(_safeTransfer(_token0, to, amount0) == false) {
```
OliverNChalk: style

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -167,8 +167,28 @@ contract ConstantProductPair is ReservoirPair {
...
+            uint256 _token0Managed = token0Managed; // gas savings
+            if (_reserve0 - _token0Managed < amount0) {
+                assetManager.returnAsset(_token0, amount0 - (_reserve0 - _token0Managed));
+                require(_safeTransfer(_token0, to, amount0), "CP: TRANSFER_FAILED");
+            }
+            else {
+                revert("CP: TRANSFER_FAILED");
+            }
+        }
+        if(_safeTransfer(_token1, to, amount1) == false) {
```
OliverNChalk: style

src/asset-management/AaveManager.sol
OliverNChalk: Ahh, my bad `msg.sender` I see.

src/asset-management/AaveManager.sol
```diff
@@ -168,6 +168,15 @@ contract AaveManager is IAssetManager, Ownable, ReentrancyGuard
         _adjustManagement(lPair, lAmount0Change, lAmount1Change);
     }
 
+    function returnAsset(address aToken, uint256 aAmount) external {
+        require(aAmount > 0, "AM: ZERO_AMOUNT_REQUESTED");
+        IAssetManagedPair lPair = IAssetManagedPair(msg.sender);
+        int256 lAmount0Change = -int256(aToken == lPair.token0() ? aAmount : 0);
+        int256 lAmount1Change = -int256(lAmount0Change == 0 ? aAmount: 0);
+        assert(lAmount0Change <= 0 && lAmount1Change <= 0);
```
OliverNChalk: Can we add a task to turn the SMTSolver on so these assertions can be proven

src/asset-management/AaveManager.sol
```diff
@@ -168,6 +168,15 @@ contract AaveManager is IAssetManager, Ownable, ReentrancyGuard
         _adjustManagement(lPair, lAmount0Change, lAmount1Change);
     }
 
+    function returnAsset(address aToken, uint256 aAmount) external {
```
OliverNChalk:
```suggestion
    // PERF: Check gas savings by specifying `address aToken, bool aToken0, uint256 aAmount`
    function returnAsset(address aToken, uint256 aAmount) external {
```

src/asset-management/AaveManager.sol
```diff
@@ -70,11 +70,11 @@ contract AaveManager is IAssetManager, Ownable, ReentrancyGuard
         IAssetManagedPair aPair,
         int256 aAmount0Change,
         int256 aAmount1Change
-    ) external nonReentrant onlyOwner {
+    ) external onlyOwner {
         _adjustManagement(aPair, aAmount0Change, aAmount1Change);
     }
 
-    function _adjustManagement(IAssetManagedPair aPair, int256 aAmount0Change, int256 aAmount1Change) private {
+    function _adjustManagement(IAssetManagedPair aPair, int256 aAmount0Change, int256 aAmount1Change) nonReentrant private {
```
OliverNChalk: lint

src/asset-management/AssetManagedPair.sol
OliverNChalk, line 136: This broke when called midswap right? Any consequence/downside of removing this update?

src/curve/stable/StablePair.sol
```diff
@@ -241,8 +260,18 @@ contract StablePair is ReservoirPair {
             }
         }
 
-        // optimistically transfers tokens
-        _safeTransfer(tokenOut, to, amountOut);
+        // if transfer fails for whatever reason
+        if(_safeTransfer(tokenOut, to, amountOut) == false) {
```
OliverNChalk: style

src/curve/stable/StablePair.sol
```diff
@@ -188,8 +188,27 @@ contract StablePair is ReservoirPair {
         amount1 = (liquidity * _reserve1) / _totalSupply;
 
         _burn(address(this), liquidity);
-        _safeTransfer(token0, to, amount0);
-        _safeTransfer(token1, to, amount1);
+
+        if(_safeTransfer(token0, to, amount0) == false) {
```
OliverNChalk: style

src/curve/stable/StablePair.sol
```diff
@@ -188,8 +188,27 @@ contract StablePair is ReservoirPair {
...
+            uint256 _token0Managed = token0Managed; // gas savings
+            if (_reserve0 - _token0Managed < amount0) {
+                assetManager.returnAsset(token0, amount0 - (_reserve0 - _token0Managed));
+                require(_safeTransfer(token0, to, amount0), "CP: TRANSFER_FAILED");
+            }
+            else {
+                revert("CP: TRANSFER_FAILED");
+            }
+        }
+        if(_safeTransfer(token1, to, amount1) == false) {
```
OliverNChalk: style

src/asset-management/AssetManagedPair.sol
xenide: didn't break. It was a conscious decision to remove it. See the very first comment I made [here](https://github.com/vexchange/v3-core/pull/96#discussion_r1005089726)

src/curve/constant-product/ConstantProductPair.sol
xenide: you mean make it like this? 
```solidity
if(!_safeTransfer(_token0, to, amount0)) 
```

Actually it was a conscious decision to use `== false` because somehow I think it's really hard to see the exclamation mark in between the `(` and the `_`

src/asset-management/AaveManager.sol
xenide: Okay. captured as issue #97

src/asset-management/AaveManager.sol
xenide: Fixed. Oh yeah wanted to highlight this. I think it's more appropriate to move the reentrancy guard to this internal function as it can be called by multiple external function and is the "heart" of the asset manager. Whereas the external `adjustManagement` function is only just a shell

src/curve/constant-product/ConstantProductPair.sol
xenide: fixed all

src/asset-management/AssetManagedPair.sol
OliverNChalk: Unless PNL has been accrued but that should be handled separately, so this seems good

src/asset-management/AaveManager.sol
OliverNChalk: Makes sense, so long as its never called twice legitimately (doesnt seem like that would ever make sense?)

src/curve/stable/StablePair.sol
```diff
@@ -241,8 +260,18 @@ contract StablePair is ReservoirPair {
             }
         }
 
-        // optimistically transfers tokens
-        _safeTransfer(tokenOut, to, amountOut);
+        // if transfer fails for whatever reason
+        if(!_safeTransfer(tokenOut, to, amountOut)) {
```
OliverNChalk: spacing

xenide: fixed

src/curve/stable/StablePair.sol
```diff
@@ -188,8 +188,27 @@ contract StablePair is ReservoirPair {
...
-        _safeTransfer(token1, to, amount1);
+
+        if (!_safeTransfer(token0, to, amount0)) {
+            uint256 _token0Managed = token0Managed; // gas savings
+            if (_reserve0 - _token0Managed < amount0) {
+                assetManager.returnAsset(token0, amount0 - (_reserve0 - _token0Managed));
+                require(_safeTransfer(token0, to, amount0), "CP: TRANSFER_FAILED");
+            }
+            else {
+                revert("CP: TRANSFER_FAILED");
```
OliverNChalk: CP -> SP

test/integration/Aave.t.sol
```diff
@@ -169,6 +169,11 @@ contract AaveIntegrationTest is BaseTest
         _increaseManagementOneToken();
     }
 
+    function testAdjustManagement_IncreaseManagementOneToken_Frozen() public
+    {
+
```
OliverNChalk, line 220: TODO

test/integration/Aave.t.sol
```diff
@@ -164,9 +185,44 @@ contract AaveIntegrationTest is BaseTest
...
+        (address lAaveToken, ,) = _dataProvider.getReserveTokensAddresses(USDC);
+        assertEq(_pair.token0Managed(), 0);
+        assertEq(_pair.token1Managed(), 0);
+        assertEq(IERC20(USDC).balanceOf(address(_pair)), MINT_AMOUNT);
+        assertEq(IERC20(lAaveToken).balanceOf(address(_manager)), 0);
+        assertEq(_manager.shares(_pair, USDC), 0);
+        assertEq(_manager.totalShares(lAaveToken), 0);
+    }
+
+    function testAdjustManagement_IncreaseManagementOneToken_Paused() public allNetworks allPairs
```
OliverNChalk, line 212: testDecrease_Paused
testDecrease_Frozen

xenide: yeah I wrote those test cases already, haven't pushed them. You can chill first, I'll let you know when I need you to review the PR again.

src/curve/stable/StablePair.sol
xenide: fixed

src/asset-management/AaveManager.sol
xenide: yup, can't think of a scenario where this is possible

src/ReservoirPair.sol
```diff
@@ -11,5 +11,48 @@ abstract contract ReservoirPair is AssetManagedPair, OracleWriter, ReentrancyGua
...
+        uint256 _reserve0 = reserve0; // gas savings
+        uint256 _reserve1 = reserve1;
+
+        _returnAndTransfer(token0, to, _totalToken0() - _reserve0, _reserve0, _reserve1);
+        _returnAndTransfer(token1, to, _totalToken1() - _reserve1, _reserve0, _reserve1);
+    }
+
+    function _returnAndTransfer(
+        address aToken, address aDestination, uint256 aAmount, uint256 aReserve0, uint256 aReserve1
+    ) internal {
```
OliverNChalk: I think we should strongly consider adopting forge fmt at some point, we have quite a few styles floating around in the codebase

src/ReservoirPair.sol
```diff
@@ -11,5 +11,48 @@ abstract contract ReservoirPair is AssetManagedPair, OracleWriter, ReentrancyGua
...
+
+        uint32 blockTimestamp = uint32(block.timestamp % 2**32);
+        uint32 timeElapsed;
+        unchecked {
+            timeElapsed = blockTimestamp - blockTimestampLast; // overflow is desired
+        }
+
+        if (timeElapsed > 0 && _reserve0 != 0 && _reserve1 != 0) {
+            _updateOracle(_reserve0, _reserve1, timeElapsed, blockTimestampLast);
+        }
```
OliverNChalk:
```suggestion
        if (timeElapsed > 0 && _reserve0 != 0 && _reserve1 != 0) {
            _updateOracle(_reserve0, _reserve1, timeElapsed, blockTimestampLast);
        }

```

src/asset-management/AaveManager.sol
```diff
@@ -96,39 +98,72 @@ contract AaveManager is IAssetManager, Ownable, ReentrancyGuard
...
+        // if the supply fails for whatever reason
+        // pool could be frozen, paused, reached its supply limit etc
+        catch {
+            // approve the token for returning to the pair
+            aToken.approve(address(aPair), aAmount);
+            rFail = true;
+            // undo shares change
+            shares[aPair][address(aToken)] -= lShares;
+            totalShares[aAaveToken] -= lShares;
+        }
```
OliverNChalk: I think all these changes are in the wrong PR? They are not directly related to `returnAsset` right? Can we remove them from this PR?

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -167,8 +148,10 @@ contract ConstantProductPair is ReservoirPair {
         amount1 = liquidity * balance1 / _totalSupply; // using balances ensures pro-rata distribution
         require(amount0 > 0 && amount1 > 0, "CP: INSUFFICIENT_LIQ_BURNED");
         _burn(address(this), liquidity);
-        _safeTransfer(_token0, to, amount0);
-        _safeTransfer(_token1, to, amount1);
+
+        _returnAndTransfer(_token0, to, amount0, _reserve0, _reserve1);
+        _returnAndTransfer(_token1, to, amount1, _reserve0, _reserve1);
```
OliverNChalk: Sounds like function transfer, this naming confused me a lot. Suggest changing it to something else. Could do `_checkedTransfer`

OliverNChalk (changes requested):

src/curve/constant-product/ConstantProductPair.sol
xenide: okay fixed

src/asset-management/AaveManager.sol
```diff
@@ -96,39 +98,38 @@ contract AaveManager is IAssetManager, Ownable, ReentrancyGuard
...
         }
         if (aAmount1Change > 0) {
-            _doInvest(aPair, lToken1, uint256(aAmount1Change));
+            _doInvest(aPair, lToken1, lToken1AToken, uint256(aAmount1Change));
         }
     }
 
-    function _doDivest(IAssetManagedPair aPair, IERC20 aToken, uint256 aAmount) private {
-        uint256 lShares = _updateShares(aPair, address(aToken), aAmount, false);
+    function _doDivest(IAssetManagedPair aPair, IERC20 aToken, address aAaveToken, uint256 aAmount) private {
```
xenide, line 119: this is for perf purposes. To save another external call when we have already queried the address of the AaveToken

xenide:
I think I managed to make it work. I analyzed the scenarios where we need / don’t need everything to revert. 

1. `adjustManagement` call by the owner
    1. this can be allowed to revert if anything fails, since it’s a standalone action made by the owner (us) 
2. `returnAsset` call by the pair
    1. this can be allowed to revert if anything fails, cuz if the asset manager can’t return enough assets to satisfy the shortfall from `swap` or `burn`, the call will fail anyway
3. `afterLiquidityEvent` call by the pair
    1. this is where the revert needs to be caught. When people mint or burn, with all other conditions valid, the mint and burn should succeed. 
    2. so this can be cleanly managed with a `try/catch` when we call it in `AssetManagedPair`

Thinking about bringing in AAVE’s error library, so that we can do `vm.expectRevert` with the specific error we are expecting. But that file's license is BUSL. What do you think? If we recreate 3/4 lines in our test file that’s not a violation right?

src/ReservoirPair.sol
xenide: I'm having trouble seeing what's the exact changes...

src/ReservoirPair.sol
xenide: okay honestly I'm only doing this to save $$$ since the audit is calculated by line....

src/ReservoirPair.sol
OliverNChalk: More logical grouping

src/asset-management/AaveManager.sol
```diff
@@ -193,19 +205,20 @@ contract AaveManager is IAssetManager, Ownable, ReentrancyGuard
...
-        rExchangeRate = IERC20(aAaveToken).balanceOf(address(this)) * 1e18 / totalShares[aAaveToken];
+        rExchangeRate = IERC20(aAaveToken).balanceOf(address(this)).divWadDown(totalShares[aAaveToken]);
     }
 
-    function _updateShares(IAssetManagedPair aPair, address aToken, uint256 aAmount, bool increase) private returns (uint256 rShares) {
-        address lAaveToken = _getATokenAddress(aToken);
-        rShares = aAmount * 1e18 / _getExchangeRate(lAaveToken);
+    function _updateShares(
+        IAssetManagedPair aPair, address aToken, address aAaveToken, uint256 aAmount, bool increase
+    ) private returns (uint256 rShares) {
```
OliverNChalk: Thoughts on removing `bool increase` seems to violate clean code principles: https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

OliverNChalk (approved):

src/asset-management/AaveManager.sol
xenide: okay makes sense. created two specific functions to take its place. 

fixed

src/ReservoirPair.sol
OliverNChalk: oh lol haha, forgot about that

PR #99, opened 2022-10-29
fix: `StablePair`::burn

xenide:
## Motivation

- make `burn` not use `try/catch` anymore since we have fixed the convergence issue
- there seems to be an argument for disallowing zero burn. 
  - see the `testBurn_DiffDecimalPlaces` test case. In cases where the 2 token decimals are not the same, it is possible that we get a non-zero amount for one token and a zero amount for the other token. The user doing this action would be doing it _to his detriment_
  - this happens because `StablePair` normalizes all token amounts to 18 decimal places for LP token calculation
  - a similar thing can also happen in `ConstantProductPair` when the differences are too big and the burn amount is too small. That's why they have a check to make sure that both tokens are `> 0` before proceeding 

## Solution

- if we disallow zero burn, it is for the benefit of the user, to make them not lose out / make this stupid mistake
- currently we have 3 options: 
   1. disallow zero burn on both pairs 
   2. allow on both pairs 
   3. status quo (allow on stable, disallow on ConstantProduct)
- my desire for uniformity of behavior across pairs pushes me to adopt option 1 or option 2. The side effects of a zero burn are just these and I don't see any problems with it: 
   1. syncing managed assets (if any)
   2. minting platform fee (if any)
   3. updating oracle 
   4. rebalance with the asset manager 
- from the gas perspective, it'd make sense for option 2 cuz it would make it cheaper for everyone by getting rid of the `require`


OliverNChalk:
> see the testBurn_DiffDecimalPlaces test case. In cases where the 2 token decimals are not the same, it is possible that we get a non-zero amount for one token and a zero amount for the other token. The user doing this action would be doing it to his detriment

Wouldn't it be more correct to say that integers round by flooring, so if a user takes out <1 of A, then they will get 0 of A. That doesn't seem like a problem to me. What if the user just wants to get rid of some dust in their wallet and arent worried about 1e-6 USDC they didnt get.

I propose doing `ii`

OliverNChalk (approved):

PR #100, opened 2022-10-31
test: demonstrate the 2 blocks in a row attack

### LINKED ISSUE #93: fix?(oracle): take into account new ETH consensus protocol and possible attacks
After the merge on ETH, there is a possibility that a validator get many blocks in a row. As seen [here](https://www.crowdfundinsider.com/2022/09/195853-euler-finance-integrates-chainlink-price-feeds-to-get-ready-for-the-merge/), [here](https://chainsecurity.com/oracle-manipulation-after-merge/), and [here](https://uniswap.org/blog/what-to-know-pre-merge). So we need to explore if there are ways to mitigate / eradicate this issue. 

Reference materials: 

1. Euler's paper on the cost of manipulating the UniV3 [price oracle](https://github.com/euler-xyz/uni-v3-twap-manipulation/blob/master/cost-of-attack.pdf) 
1. Euler finance's [median oracle](https://github.com/euler-xyz/median-oracle) solution
1. [oracle attack simulator](https://www.euler.finance/blog/uniswap-oracle-attack-simulator) 

Possible solutions:

1. Clamping the rate of change (see comment below)
2. Increasing the window with which the average TWAP is calculated. E.g. instead of using a 5 min window, use a 1/2 hour window
   - lengthening the window increases the cost of attack 
4. Do not use our own AMM oracle on applicable chains (ETH mainnet, polygon), and instead rely on 3rd party solutions like `chainlink`

Attack PoC: 
- [here](https://github.com/vexchange/v3-core/commit/94654fe82f16154d6fdd586dc884b5258bff689a)

xenide:
## Motivation

- don't necessarily have to merge this but a PR is a good format to discuss and comment on code
- wrote a test case to demonstrate the attack before trying any solutions 
- Conditions and assumptions
   - ConstantProductPair for ETH-USDC with $3M TVL, swap fee 0.3%
      - UniV2's USDC-ETH pool has 95M locked while UniV3 USDC-ETH pool has 195M locked. A little too much for our simulations
      - Our pools would probably not reach that size so it makes sense to model with something more modest
   - Attacker controls X consecutive blocks, no more than that 

- currently it shows that with a window period of 60 mins and attacker controlling 5 blocks: 
   - it costs: 3.02 ETH + 4500 USDC for the attacker (input of `12_000_000e18` ETH)
      - somehow I don't know why the cost is so low 
      - it's because it's 0.3% * 1000 ETH (initial amount) * 2
   - moves the instantaneous price by:  `1e8`  from (1500 USDC to 0.0000104 USDC )
   - moves the TWAP by: -22%  (average price over 60min is 1168 USD)

- with a window period of 60 mins and attacker controlling 10 blocks
  - with the same costs and price changes
  - moves the TWAP by -43% (average price over 60min is 855 USD) 

- with a window period of 15 mins and attacker controlling 5 blocks
  - with the same costs and price changes
  - moves TWAP by -62.7% (average price over 15 mins is 558 USD)

- with a window period of 15 mins and attacker controlling 10 blocks
  - with the same costs and price changes
  - moves TWAP by -89.2% (average price over 15 mins is 162 USD)
  

## Solution

- not sure how to proceed from here. We could go in several directions 
  1. make cases for different window periods (current one is 60 mins), and see how the window period affects the amount needed to manipulate the oracle
  2. For the test cases we can fix the magnitude we want to manipulate by. (e.g. by depressing the price 90% or inflating the price 10X), how much does it cost in all these varying conditions

### DIFF
--- a/test/unit/ConstantProductPair.t.sol
+++ b/test/unit/ConstantProductPair.t.sol
@@ -558,7 +558,7 @@ contract ConstantProductPairTest is BaseTest
 
 
     ///     ORACLE ATTACK TEST CASES
-    function testOracleAttack2Blocks() public
+    function testOracleAttack5Blocks() public
     {
         // Create a ETH-USDC pair
         MintableERC20 lWEth = new MintableERC20("WETH", "WETH", 18);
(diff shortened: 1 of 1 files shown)


OliverNChalk:
Input ETH amount is multiples of the circulating supply though? The problem with a 2 block attack is one cannot flash loan/mint the amount. Might be worth doing a 5 & 10 block attack but with more realistically achievable numbers

xenide:
Closing as the impact of the attack is understood

PR #104, opened 2022-11-01
feat: clamped oracle 

### LINKED ISSUE #103: feat: implement clamped oracle alongside the unclamped one
As a proof of concept and to address #93, implement this with configured parameters for: 

1. max rate of change per unit time

xenide:
## Solution
- in terms of storing the clamped price, I decided to use another `uint`, so that every time the oracle needs to be updated, the `prevClampedPrice` can be read. This is the the most gas efficient solution I can think of, which costs 1 additional `SLOAD` + an `SSTORE`. The alternative is reading 2 previous oracle observations to determine the price in-between those two time periods, and that costs 2 `SLOAD`s

- in terms of sizes for the struct, I was thinking the raw accumulators are going to be more useful / real in the long run, so It makes sense to give it more space 
  - but honestly given the time it takes to overflow, this doesn't matter much
  - but if we wanted symmetric sizes, we could do `uint84` for both price accmulators, and `uint56` for the liquidity accumulator


src/ReservoirPair.sol
```diff
@@ -55,4 +55,9 @@ abstract contract ReservoirPair is AssetManagedPair, OracleWriter, ReentrancyGua
         blockTimestampLast = blockTimestamp;
         emit Sync(reserve0, reserve1);
     }
+
+    function setMaxChangePerSecond(uint8 aChangePerSecond) external override onlyFactory {
+        require(0 <= aChangePerSecond && aChangePerSecond <= MAX_CHANGE_PER_SEC, "RP: INVALID_CHANGE_PER_SECOND");
+        allowedChangePerSecond = aChangePerSecond;
```
OliverNChalk: should be `_allowedChangePerSecond` if its private/internal. That said, shouldnt this be public?

src/interfaces/IOracleWriter.sol
```diff
@@ -2,14 +2,19 @@ pragma solidity 0.8.13;
...
 interface IOracleWriter {
     struct Observation {
-        // natural log (ln) of the price (token1/token0)
-        int112 logAccPrice;
-        // natural log (ln) of the liquidity (sqrt(k))
-        int112 logAccLiquidity;
+        // natural log (ln) of the raw accumulated price (token1/token0)
+        int112 logAccRawPrice;
+        // natural log (ln) of the clamped accumulated price (token1/token0)
+        // even in the case of extreme prices (3e75), will overflow once every 878 years
```
OliverNChalk: Is 3e75 the extremity of the price?

OliverNChalk: Can you include 3e75 shown as 2**N so that we can see the relationship between this number and the size of liquidity in bits?

src/interfaces/IOracleWriter.sol
```diff
@@ -2,14 +2,19 @@ pragma solidity 0.8.13;
...
+        // even in the case of extreme liquidity (3e75), will overflow once every 878 years
+        int56 logAccLiquidity;
         // overflows every 136 years, in the year 2106
         uint32 timestamp;
     }
 
-    function observations(uint256 index) external view returns (int112, int112, uint32);
+    function observations(uint256 index) external view returns (int112, int56, int56, uint32);
     function index() external returns (uint16);
+    function setMaxChangePerSecond(uint8) external;
```
OliverNChalk: Is the reason for uint8 purely because the MAX < uint8.max? I would imagine this increases gas cost due to masking, given we store this as a uint64?

src/oracle/OracleWriter.sol
```diff
@@ -1,16 +1,56 @@
 pragma solidity 0.8.13;
 
-import "src/interfaces/IOracleWriter.sol";
+import { IOracleWriter } from "src/interfaces/IOracleWriter.sol";
+import { LogCompression } from "src/libraries/LogCompression.sol";
+import { stdMath } from "forge-std/Test.sol";
 
 abstract contract OracleWriter is IOracleWriter {
+    // 10 basis points per second which is 6% per minute and doubling of the price in 16 minutes
+    uint256 internal constant MAX_CHANGE_PER_SEC = 0.001e18;
```
OliverNChalk: I think this is quite a high rate of change but still, to set it as a  compile time max seems like we would want to set it even higher in case there is ever such a use case?

src/oracle/OracleWriter.sol
```diff
@@ -1,16 +1,56 @@
...
+
     Observation[65536] public observations;
     uint16 public index = type(uint16).max;
 
+    // maximum allowed rate of change of price per second
+    // for clamping the price
+    uint64 internal allowedChangePerSecond = 0.0005e18; // 5 bp
```
OliverNChalk: Feels arbitrary, shouldn't we just have a constructor argument?

test/__fixtures/BaseTest.sol
```diff
@@ -87,12 +87,13 @@ abstract contract BaseTest is Test {
...
-        int112 aPrice,
-        int112 aLiq,
+        int112 aRawPrice,
+        int56 aClampedPrice,
+        int56 aLiq,
         uint32 aTime
     ) internal
     {
-        bytes32 lEncoded = bytes32(abi.encodePacked(aTime, aLiq, aPrice));
+        bytes32 lEncoded = bytes32(abi.encodePacked(aTime, aLiq, aClampedPrice, aRawPrice));
```
OliverNChalk: There's a replacement API for `abi.encodePacked`, which is `bytes.concat`, it could be worth trying out.

test/unit/ConstantProductPair.t.sol
```diff
@@ -547,8 +549,62 @@ contract ConstantProductPairTest is BaseTest
...
+        (int112 lAccRawLogPrice0, int56 lAccClampedLogPrice0, , uint32 lTimestamp0) = _constantProductPair.observations(0);
+        (int112 lAccRawLogPrice1, int56 lAccClampedLogPrice1, , uint32 lTimestamp1) = _constantProductPair.observations(1);
+
+        // act
+        _stepTime(5);
+        _constantProductPair.sync();
+
+        // assert - make sure that the prevClampedPrice is keeping up
+        (int112 lAccRawLogPrice2, int56 lAccClampedLogPrice2, , uint32 lTimestamp2) = _constantProductPair.observations(2);
+        assertEq(_constantProductPair.prevClampedPrice(), 1.00500625e18);
```
OliverNChalk, line 584: Id be interested in asserting the the unclamped price is more extreme. Because that would show that clamping has occurred, similar to how you wrote the next test

src/oracle/OracleWriter.sol
```diff
@@ -1,16 +1,56 @@
...
+    uint256 public prevClampedPrice;
```
OliverNChalk, line 26: Seems like we need to write a whole extra word for every swap? How many more `SSTORE`s is the new branch, also what is the gas difference?

xenide: as discussed, I've made a perf note to try read the two previous observations to figure out the clamped price instead of using this new variable

test/__fixtures/BaseTest.sol
xenide: okay I attempted this. `bytes.concat` needs explicit bytes type and doesn't recognize the int types so we have to cast it manually, sometimes multiple times. Becomes pretty verbose and ugly

So I think no need to use that for now

test/unit/ConstantProductPair.t.sol
xenide: okay I added an `assert` to show that the unclamped TWAP price is a lot bigger

src/ReservoirPair.sol
xenide: I was thinking about this... if we make it public, it'd increase the gas cost of function lookups. 

The only good reason to have it is for admins like us to check that this variable is correctly set? Are there other legit uses for the world to know about this rate?

src/interfaces/IOracleWriter.sol
xenide: this was an error on my part. Have changed the arg type to `uint64`, to match the type of the storage variable. 

Even if we use `uint64`, there still the masking problem right? Should I just use `uint256` instead?

src/interfaces/IOracleWriter.sol
xenide: alright so the `LogExpMath` lib can take in really large numbers (at the limit of `int256.max` (around `5e77`)) for compressing a raw value into an exponent. But the decompression function has a limit of around `e ** (130) == 2.87e56`. Actually speaking of this, it seems a little dangerous that balancer's `LogCompression` library doesn't check the arg value before casting from `uint256` to `int256`

When I wrote the comment I wanted to show that even with values beyond the supported range, the accumulator would take too long to overflow. 

So for the sake of proper and sane documentation, we can work with the max exponent supported by the library, which is `130.0000` (fits within 21 bits, but is not a nice whole exponent of 2). Multiply that with the 32 bits of the timestamp, we can be guaranteed that everything will fit into 53 bits. Thus 56 bits for the accumulator is more than enough. Should I document it as such?

src/oracle/OracleWriter.sol
xenide: so do you propose we do not have an max absolute (constant) cap, like we do for the swap / platform fees? 

I was thinking that having MIN / MAX hard caps could prevent misconfigurations by the admins (us)

src/oracle/OracleWriter.sol
xenide: if we have a constructor argument, who provides this rate? Certainly not the user? If it's not the user, then it's a default value that's read from the factory config mapping? Reading from the factory default config is not so diff from the default I'm value I'm using here. 

My thinking was we can set a sensible rate as the default value which all pairs use. Then in production, for specific pairs that will be used as price feeds but have too loose / tight a clamp (which is probably not many), we can then tweak it.

src/ReservoirPair.sol
OliverNChalk: Yeah, I think contract auditability is worth it because it lets users know about the parameters when they look at it on Etherscan. We can always revisit this during optimization time

src/interfaces/IOracleWriter.sol
OliverNChalk: Yes please

src/interfaces/IOracleWriter.sol
OliverNChalk: If we're not packing, then I would use uint256

src/oracle/OracleWriter.sol
OliverNChalk: I can see the value in catching fat fingers but this can never be circumvented either. When we set the clamping to a really high number it doesnt insta rug it just behaves like UniV3 which is theoretically vulnerable under certain conditions.

Is there a number we can choose that protects against 2 or 3 block attacks but still gives us huge wiggle room. E.g. if 24-36 seconds is a 3 block attack then can we do like a 5% move every 36 seconds? Protects fat fingers, nullifies 3 block attacks but still gives us huge range?

OliverNChalk: I guess 5% per 36 seconds is close to 6% per minute. Maybe your value is already quite good...

src/oracle/OracleWriter.sol
OliverNChalk: `Reading from the factory default config is not so diff from the default I'm value I'm using here.` - The factory default can be edited over time so its completely different. This one is immutable.

test/__fixtures/BaseTest.sol
OliverNChalk: It calls out your struct in code right. uint64 = bytes8, uint128 = bytes16 etc. Right now the reader has no clue whether your bytes32 is going to line up with your abi.encodePacked input

OliverNChalk: My understanding is it should just be `bytes.concat(bytes4(aTime), bytesN(aLiq), bytesN(aClampedPrice), bytesN(aRawPrice))`

src/ReservoirPair.sol
xenide: yes auditability is important. Changed the visibility of the variable to `public`. 

Usually when some state of the contract changes we'd emit an event. In this case do you think it's necessary? Doesn't seem very necessary to me

src/interfaces/IOracleWriter.sol
xenide: okay fixed.

src/oracle/OracleWriter.sol
xenide: okay makes sense. Made it read from the factory on instantiation. 

Also, since `setAllowedChangePerSecond` and the constructor now need to know about the factory (previously they didn't), we have to change the inheritance a bit (make `OracleWriter` inherit from `Pair`), so that the impl of `setAllowedChangePerSecond` can stay inside `OracleWriter` instead of having to be in `ReservoirPair`.

test/__fixtures/BaseTest.sol
xenide: okay for better readability and verifiability, I've changed it. please see if that's the effect you wanted. 

Because solidity doesn't allow an explicit conversion from `int` to `bytes`, we have to convert the `int`s to`uint`s first, which the reason for the ugliness I said previously

src/oracle/OracleWriter.sol
xenide: We can set this to a pretty high number and still achieve have the desired mitigating effect. Because for the attackers to successfully manipulate the TWAP meaningfully, they have to manipulate the spot prices by A LOT during those 5 blocks. 

So imo, there are two ways to derive this threshold:
1. empirically, we can look at real DEX data, and then pick a threshold that covers 99.9% (or whatever %) of cases. Kind of like hypothesis testing in statistics
2. by feeling. If we set it to `0.01e18` (10X the current rate), then in the case of a 5-block attacker, he can move the clamped price `[12%, 24%, ... , 60%]` in 1 minute. Don't forget in the next minute it will be `[60%, 48%, ... 0]`, but that's nothing like manipulating prices 1000X usually needed. In the grand scheme of a 30min TWAP, maybe he effectively moved it 5%

src/interfaces/IOracleWriter.sol
xenide: okay changed the documentation. The docs describing the liquidity accumulator is not implemented yet, but will be done in #105

src/ReservoirPair.sol
OliverNChalk: It's pretty cheap to do given it's a setter variable. In general im not sure how I feel about events. I think eventually EVM LOG will be reworked, for now I suggest emitting an event to help 3rd party tooling that perhaps wants to monitor this stuff.

src/oracle/OracleWriter.sol
OliverNChalk: I propose 10x to keep it simple & done

src/oracle/OracleWriter.sol
OliverNChalk: okay

test/__fixtures/BaseTest.sol
OliverNChalk: Isn't the conversion unsafe though? `int` -> `uint` -> `bytes` or does solidity not overflow check by default?

xenide: when we do an explicit conversion in solidity, no overflow check is done. Solidity kind of assumes we know what the hell we're doing when we use explicit conversions

Also I don't know of any other way other than the `int` -> `uint` -> `bytes` path. My research also shows the same

src/oracle/OracleWriter.sol
xenide: okay I agree. done.

src/ReservoirPair.sol
xenide: okay implemented with test. Not emitting event at instantiation but when changed by the factory.

OliverNChalk: Why not at instantiation?

src/oracle/OracleWriter.sol
```diff
@@ -1,11 +1,69 @@
...
 
+    // maximum allowed rate of change of price per second
+    // to mitigate oracle manipulation attacks in the face of post-merge ETH
+    uint256 public allowedChangePerSecond;
+    uint256 public prevClampedPrice;
+
+    constructor() {
+        uint256 lAllowedChangePerSecond = factory.read("Shared::allowedChangePerSecond").toUint256();
+        require(0 < lAllowedChangePerSecond && lAllowedChangePerSecond <= MAX_CHANGE_PER_SEC, "OW: INVALID_CHANGE_PER_SECOND");
+        allowedChangePerSecond = lAllowedChangePerSecond;
```
OliverNChalk:
```suggestion
        uint256 lAllowedChangePerSecond = factory.read("Shared::allowedChangePerSecond").toUint256();
        setAllowedChangePerSecond(lAllowedChangePerSecond);
```

src/oracle/OracleWriter.sol
```diff
@@ -1,11 +1,69 @@
...
+        }
+        else {
+            rClampedPrice = aCurrRawPrice;
+            rClampedLogPrice = int112(LogCompression.toLowResLog(aCurrRawPrice));
+        }
+    }
+
+    function _calcPercentageDiff(uint256 a, uint256 b) private pure returns (uint256) {
+        return stdMath.percentDelta(a, b);
+    }
```
OliverNChalk, line 65: Can you make a test case that tests the threshold +- 1 so that we can detect if `percentDelta` ever changes behaviour?

OliverNChalk (approved):

src/ReservoirPair.sol
xenide: okay so, I'm aware that it's the convention to emit events even for instantiation. e.g. `Ownable` emits `OwnershipTransferred` when the contract is created. 

But currently we don't emit events it for `swapFee` and `platformFee` on instantiation. Should we change that as well? If we use the method you're proposing below, we can just call `updateSwapFee` and `updatePlatformFee` in the constructor, which simplifies things.

src/oracle/OracleWriter.sol
xenide: fixed

src/ReservoirPair.sol
OliverNChalk: Yes we should change those as well in a separate issue

src/oracle/OracleWriter.sol
xenide: okay created a few basic tests. Only with the `uint` function since that's what we are using.

Also discovered that the order of arguments matters. The base of comparison should be `b` which is what we are doing

src/ReservoirPair.sol
xenide: okay created #109 to be addressed with a PR later

src/oracle/OracleWriter.sol
OliverNChalk: Those are good but I also meant tests of the oracle writer tolerance. So that regardless of what internal implementation we use we have locked in the observable behavior to be how we expect/want it to be.

xenide:
the legacy tests have been failing since #[8cef5e5](https://github.com/vexchange/v3-core/pull/104/commits/8cef5e5dabcec57ff3fc84e752a8b5d66249ea59), I've spent hours trying to figure out why and tried many things.  

it's failing when `factory.addCurve` [function](https://github.com/vexchange/uniswap-v2-core/blob/f849f02bcb3a032e77735affc97549e161dd32a0/test/shared/fixtures.ts#L42) calls the `SSTORE2::write` library and that's exactly where it fails. 

Right now I think the most likely cause is that, the bytecode is too large (now it's more than 24K, previously it was under), and the legacy waffle / ganache / evm hard fork target versions don't support it. But I can't 100% confirm this. It's been working fine for a long without issues. And when I tried adding curves with truncated bytecodes it works. 

I'm going ahead to merge this PR first, and we can fix the legacy tests later on.

src/oracle/OracleWriter.sol
xenide: okay I wrote two cases at the edges. Please see if they're sound

test/unit/ConstantProductPair.t.sol
```diff
@@ -608,4 +608,49 @@ contract ConstantProductPairTest is BaseTest
...
+        assertEq(_constantProductPair.prevClampedPrice(), 1e18);
+
+        // act
+        _stepTime(5);
+        _constantProductPair.sync();
+        console.log(_constantProductPair.prevClampedPrice());
+        // assert
+        (int112 lAccRawLogPrice1, int56 lAccClampedLogPrice1, , uint32 lTimestamp1) = _constantProductPair.observations(1);
+        // no diff between raw and clamped prices
+        assertEq(_constantProductPair.prevClampedPrice(), 1.0025e18);
```
OliverNChalk: Shouldn't we be checking that lAccRawLog != lAccClamped, otherwise it feels like this test isn't asserting anything right?

OliverNChalk (approved):

test/unit/ConstantProductPair.t.sol
xenide: so in this case difference is so small, it's between the margin of error for the log compression algorithm. 

For the `prevClampedPrice` since it's precise, there is a difference between at limit and over limit. For the accumulator, it's the same, cuz the raw-clamped price diff has not moved more than the sensitivity of our log compression algorithm. If we want, we can increase the swap amount such that it moves the needle, but it won't be a +- 1 kind of thing but slightly coarser

OliverNChalk: yes please, i think increase the swap amount is good

OliverNChalk: Just so long as the two test cases show a difference that we understand its fine

PR #105, opened 2022-11-02
fix: liq calculation in oracle for `StablePair`

### LINKED ISSUE #102: test: show that the liquidity oracle of the two curves are comparable
For ConstantProduct and Stable, the way liquidity is calculated is slightly different. But we should be able to compare them in an equal / fair fashion such that when the consumer of the oracle reads the two values, it is able to choose the one with greater liquidity.

xenide:
## Motivation
- while writing the test cases, I discovered an issue: currently the way we are calculating liquidity for the ConstantProductPair, we just do `x * y`. This doesn't take into account the decimal places of the tokens involved (for oracle price calc we already do). But for StablePair, it takes into account the decimals. As such, the numbers will differ greatly (by orders of magnitude for coins with different decimal places). 

- we need a way to fairly compare the liquidity in ConstantProduct and Stable pairs 

## Solution
- we should normalize the token balances to 18 when calculating the liquidity as well
  - currently we already do this for the price so it's just extending it to liquidity as well 
  - fix is trivial, but might be made irrelevant with the problem below

- to compare the liquidity of diff curves, honestly this is difficult 
  - I can't think of a way to massage the numbers such that they're comparable
    - both addition and multiplication are "one way" functions, we can't get its constituents just from the sum/product even if we only need one token's liquidity 

## Workarounds if we cannot find a solution
1. for the consumer, only compare liquidity across same type of curve, not diff types of curves
   - I guess this is not really an acceptable solution is it?  
3. use the stable invariant calculation on the ConstantProduct reserves, or use the `liq = x * y`  on the StablePair 
4. simply summing the two reserves together
   - but one day when one asset depegs / goes to shit the liq would just very high on that pair. The pair has a lot of liquidity, just that it's shit liquidity <== not sure if this argument is valid 
5. only track liquidity of token0 
   - since the oracle consumer is interested in selecting which ReservoirPair with the same token0+token1 (but diff curves) to trust more with the prices, and assuming that any difference between the two are arbed by market participants, simply tracking the reserves of ONE token is a good enough proxy for that purpose 

it seems that among these options, option 2 (`x * y`) and option 4 makes the most sense 

by doing this we're acknowledging that the use case for the oracle is:
- primarily to determine which of the ConstantProductPair / StablePair has more liquidity, and thus to trust for the oracle prices and that
- the liquidity captured by the oracle is not indicative of the `totalSupply` of the LP tokens


OliverNChalk:
Option 4 is a lazy version of 3, so if you dont like 3 then you must not like 4 by analogy, right?

That said, I agree with you option 2 seems reasonable. It is essentially option 4 but it is:

1. A smaller number than summing (resistant to overflow)
2. Less divergence from UniV2 which was quite an elegant protocol

To confirm this liquidity number will be stored compressed? Because it could be uint112 if we have maxed 0 and 1 reserves

xenide:
Yes this will be stored compressed, using the same balancer lib as we do for the price etc. 

At the max of `uint112 * uint112` (`uint224`), the resulting log of that is `155`, which is larger than the `130` limit set by the balancer library. The writing can continue with no problem, but when the consumer of the oracle tries to calculate `e^155`, the lib will throw an error. 

How do we reconcile this? One way I can think of is to take the sq root of `x * y`. That will definitely be within the range supported. `log_e(sqrt(uint224)) => 77.6325`, which is `< 130`.

Also, since we're just doing `x * y`, there is no need to normalize the token amounts into 18 decimals.

OliverNChalk:
Oh, I always thought it was sqrt(x*y) like UniV2 liquidity growth

OliverNChalk:
If you don't do sqrt then liquidity scales N^2 instead of N. We want to compare this liqs so its good for them to scale fairly. e.g. 100 A 100 B = 10,000; 200 A 200 B = 40,000. sqrt(10,000) = 100; sqrt(40,000) = 200 so scales linearly

xenide:
Re-titled the PR to fixing the implementation of the liquidity oracle 

Since we have no need to show that the liquidity recorded on on both curves are comparable now (they’re literally the same impl), I’ve written some scenarios to show how both curves can end up providing:
1. the same prices with different reserves, and 
2. the same price with the same liq. 

It's important / interesting for us to note that: 

1. Since our oracle liq is the geometric mean of the two reserves, the when the pools are imbalanced, it's more likely to see a higher liq value from `CP` compared to `SP`.
   - the more unbalanced the pools are, the more pronounced this effect is 
   - at perfect balance (1:1), the difference is 0

OliverNChalk (approved):

PR #112, opened 2022-11-15
feat: oracle reader access control

xenide:
# WIP 

## Solution

- currently we have the following roles. Do you think it makes sense to refactor and use an auth library? 
1. `factory`
2. `assetManager`
3. `authorizedOracleCaller`

- in this architecture there will be no "custom" oracle callers, only the one that is specified by the factory


src/curve/stable/StablePair.sol
```diff
@@ -11,7 +11,7 @@ import { GenericFactory } from "src/GenericFactory.sol";
 import { IReservoirCallee } from "src/interfaces/IReservoirCallee.sol";
 import { StableMath } from "src/libraries/StableMath.sol";
 import { StableOracleMath } from "src/libraries/StableOracleMath.sol";
-import { ReservoirPair } from "src/ReservoirPair.sol";
+import { ReservoirPair, Observation } from "src/ReservoirPair.sol";
```
OliverNChalk, line 14: Interesting, is it necessary to re-declare the Observation import all the time?

src/oracle/OracleWriter.sol
```diff
@@ -17,19 +17,34 @@ abstract contract OracleWriter is Pair, IOracleWriter {
     uint256 internal constant MAX_CHANGE_PER_SEC = 0.01e18;
     string internal constant ALLOWED_CHANGE_NAME = "Shared::allowedChangePerSecond";
 
-    Observation[65536] public observations;
+    Observation[65536] internal _observations;
     uint16 public index = type(uint16).max;
 
+    mapping(address => bool) authorizedCallers;
```
OliverNChalk: I don't like using a mapping, can we just have a single address called `OracleReader` that is authorized? That address can then be changed to support contract upgrading.

In general we should model our system like MakerDAO does, small components that can be swapped out by simply changing an address.

src/oracle/OracleWriter.sol
```diff
@@ -17,19 +17,34 @@ abstract contract OracleWriter is Pair, IOracleWriter {
...
         setAllowedChangePerSecond(lAllowedChangePerSecond);
     }
 
+    function observations(uint256 aIndex) external view authorized returns(Observation memory rObservation) {
+        return _observations[aIndex];
+    }
+
+    function setAuthorized(address aCaller, bool aAuthorized) external onlyFactory {
+        authorizedCallers[aCaller] = aAuthorized;
+    }
```
OliverNChalk: When you move array from a mapping model, you should then support pulling the new address from the factory, similar to fees

OliverNChalk: Also, does this design mean we need to configure every single pair to be authorized? Seems very intense

OliverNChalk (changes requested):

src/curve/stable/StablePair.sol
xenide: yes seems to be the case if we want to use the struct

src/oracle/OracleCaller.sol
```diff
@@ -0,0 +1,24 @@
...
+
+    constructor(address aOwner) Owned(aOwner) {} // solhint-disable-line no-empty-blocks
+
+    function observation(IOracleWriter aPair, uint256 aIndex) external view returns (Observation memory rObservation) {
+        require(whitelist[msg.sender], "OC: NOT_WHITELISTED");
+        rObservation = aPair.observation(aIndex);
+    }
+
+    function whitelistAddress(address aCaller, bool aWhitelist) external onlyOwner {
+        if (whitelist[aCaller] != aWhitelist) {
```
OliverNChalk: I would remove this check, increases gas cost for smart humans and only save gas for silly humans (with no loss of funds possible)

src/oracle/OracleWriter.sol
```diff
@@ -16,18 +16,38 @@ abstract contract OracleWriter is Pair, IOracleWriter {
     // 100 basis points per second which is 60% per minute
     uint256 internal constant MAX_CHANGE_PER_SEC = 0.01e18;
     string internal constant ALLOWED_CHANGE_NAME = "Shared::allowedChangePerSecond";
+    string internal constant ORACLE_CALLER_NAME = "Shared::oracleCaller";
 
-    Observation[65536] public observations;
+    Observation[65536] public _observations;
```
OliverNChalk: Should be private right?
```suggestion
    Observation[65536] private _observations;
```

src/oracle/OracleWriter.sol
```diff
@@ -16,18 +16,38 @@ abstract contract OracleWriter is Pair, IOracleWriter {
...
+        rObservation = _observations[aIndex];
+    }
+
+    function updateOracleCaller() public {
+        address lNewCaller = factory.read(ORACLE_CALLER_NAME).toAddress();
+        if (lNewCaller != oracleCaller) {
+            emit OracleCallerChanged(oracleCaller, lNewCaller);
+            oracleCaller = lNewCaller;
+        }
     }
```
OliverNChalk, line 47: Here the check is fine to avoid spammers

src/oracle/OracleWriter.sol
```diff
@@ -16,18 +16,38 @@ abstract contract OracleWriter is Pair, IOracleWriter {
...
     // to mitigate oracle manipulation attacks in the face of post-merge ETH
     uint256 public allowedChangePerSecond;
     uint256 public prevClampedPrice;
 
+    address public oracleCaller;
+
+    modifier onlyOracleCaller() {
+        require(msg.sender == oracleCaller, "OW: NOT_ORACLE_CALLER");
+        _;
+    }
```
OliverNChalk: I think its more readable & less code to just inline this check, agree? Vyper actually does not support modifiers btw

OliverNChalk (commented):
Looking good, just some minor comments and what looks like 1 bug

src/oracle/OracleWriter.sol
xenide: that's right. fixed

src/oracle/OracleCaller.sol
xenide: alright fixed

src/oracle/OracleWriter.sol
xenide: yeah I thought about it too. It is indeed fewer lines of code to inline it (which saves us money). 

Btw how is vyper relevant to our code here? We ain't using vyper

OliverNChalk: One should study history and the approach of others when attempting to determine the best path for themselves

OliverNChalk (approved):

PR #119, opened 2022-11-21
perf: use gas efficient libraries

### LINKED ISSUE #62: optimization: use gas efficient libraries
Some of our code uses openzeppelin which might not be the most gas efficient out there. We should use the most gas efficient libraries out there before putting the code into production. At the time of writing `solmate` seems to be the most suitable candidate. 

## List of things to change
- [ ] ConstantProductPair::lock => solmate::ReentrancyGuard

xenide:
## What I've tried

1. replace openzeppelin's ERC20 with forge-std's IERC20 (this has the optional names, symbol, decimals functions)
   - did not make any difference at all to gas usage 
   - therefore did not push the changes to this PR 
1. replace openzeppelin's ReentrancyGuard with solmate's 
   - in this case, solmate's impl seems to be more gas efficient, with just one or two outliers
   - red is using openzeppelin's, green is using solmate's 

```diff
comparing openzeppelin's ReentrancyGuard with solmate's
104,105c104,105
< StablePairTest:testRampA_SwappingDuringRampingDown(uint256,uint64,uint64,uint128) (runs: 10000, μ: 96593, ~: 101494)
< StablePairTest:testRampA_SwappingDuringRampingUp(uint256,uint64,uint64,uint128) (runs: 10000, μ: 96736, ~: 101116)
---
> StablePairTest:testRampA_SwappingDuringRampingDown(uint256,uint64,uint64,uint128) (runs: 10000, μ: 96594, ~: 101494)
> StablePairTest:testRampA_SwappingDuringRampingUp(uint256,uint64,uint64,uint128) (runs: 10000, μ: 96735, ~: 101116)
124,126c124,126
< StableMathTest:testComputeLiquidityFromAdjustedBalances_ConvergeEvenWithVeryUnbalancedValues(uint256,uint256,uint256) (runs: 10000, μ: 19002, ~: 13720)
< StableMathTest:testGetAmountIn(uint256,uint256,uint256) (runs: 10000, μ: 24657, ~: 23987)
< StableMathTest:testGetAmountOut(uint256,uint256,uint256) (runs: 10000, μ: 27248, ~: 26263)
---
> StableMathTest:testComputeLiquidityFromAdjustedBalances_ConvergeEvenWithVeryUnbalancedValues(uint256,uint256,uint256) (runs: 10000, μ: 19004, ~: 13720)
> StableMathTest:testGetAmountIn(uint256,uint256,uint256) (runs: 10000, μ: 24656, ~: 23987)
> StableMathTest:testGetAmountOut(uint256,uint256,uint256) (runs: 10000, μ: 27250, ~: 26263)
```

3. replace solmate's `Owned` with openzeppelin's `Ownable`.
   - red is using openzeppelin's, green is using solmate's 
   - solmate's seems to use more gas inefficient, which is surprising. In most cases the gas cost increased
     - this could be due to solmate's impl having a constructor which takes in an arg 
     - besides that I was really expecting solmate's to be better cuz it's more minimalist with fewer functions

```diff
comparing ownable vs owned
< ConstantProductPairTest:testOracle_ClampedPrice_AtLimit() (gas: 226480)
< ConstantProductPairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 226449)
< ConstantProductPairTest:testOracle_ClampedPrice_OverLimit() (gas: 226982)
< ConstantProductPairTest:testOracle_CorrectLiquidity() (gas: 312171)
< ConstantProductPairTest:testOracle_CorrectPrice() (gas: 321094)
< ConstantProductPairTest:testOracle_LiquidityAtMaximum() (gas: 298513)
< ConstantProductPairTest:testOracle_NoWriteInSameTimestamp() (gas: 227350)
< ConstantProductPairTest:testOracle_OverflowAccLiquidity() (gas: 127902)
< ConstantProductPairTest:testOracle_OverflowAccPrice() (gas: 181946)
< ConstantProductPairTest:testOracle_SimplePrices() (gas: 319476)
---
> ConstantProductPairTest:testOracle_ClampedPrice_AtLimit() (gas: 226502)
> ConstantProductPairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 226471)
> ConstantProductPairTest:testOracle_ClampedPrice_OverLimit() (gas: 227004)
> ConstantProductPairTest:testOracle_CorrectLiquidity() (gas: 312215)
> ConstantProductPairTest:testOracle_CorrectPrice() (gas: 321160)
> ConstantProductPairTest:testOracle_LiquidityAtMaximum() (gas: 298557)
> ConstantProductPairTest:testOracle_NoWriteInSameTimestamp() (gas: 227372)
> ConstantProductPairTest:testOracle_OverflowAccLiquidity() (gas: 127968)
> ConstantProductPairTest:testOracle_OverflowAccPrice() (gas: 182012)
> ConstantProductPairTest:testOracle_SimplePrices() (gas: 319542)
32,46c32,46
< ConstantProductPairTest:testWriteObservations() (gas: 253019)
< GenericFactoryTest:testAddCurve() (gas: 70930)
< GenericFactoryTest:testAddCurve_OnlyOwner() (gas: 13591)
< GenericFactoryTest:testGetPair() (gas: 18131)
< OracleCallerTest:testObservation_NotWhitelisted(uint256) (runs: 256, μ: 85734, ~: 85104)
< OracleCallerTest:testWhitelistAddress() (gas: 146566)
< OracleCallerTest:testWhitelistAddress_NotOwner() (gas: 15159)
---
> ConstantProductPairTest:testWriteObservations() (gas: 253129)
> GenericFactoryTest:testAddCurve() (gas: 70906)
> GenericFactoryTest:testAddCurve_OnlyOwner() (gas: 13570)
> GenericFactoryTest:testGetPair() (gas: 18129)
> OracleCallerTest:testObservation_NotWhitelisted(uint256) (runs: 256, μ: 85800, ~: 85170)
> OracleCallerTest:testWhitelistAddress() (gas: 146606)
> OracleCallerTest:testWhitelistAddress_NotOwner() (gas: 15160)
49,52c49,52
< OracleWriterTest:testOracle_CompareLiquidityTwoCurves_Balanced() (gas: 225084)
---
> OracleWriterTest:testOracle_CompareLiquidityTwoCurves_Balanced() (gas: 225128)
56,57c56,57
< OracleWriterTest:testUpdateOracleCaller() (gas: 113283)
< OracleWriterTest:testUpdateOracleCaller_NoChange() (gas: 81333)
---
> OracleWriterTest:testUpdateOracleCaller() (gas: 113279)
> OracleWriterTest:testUpdateOracleCaller_NoChange() (gas: 81377)
62,68c62,68
< PairTest:testRecoverToken() (gas: 200579)
< PairTest:testSetPlatformFeeForPair() (gas: 110805)
< PairTest:testSetPlatformFeeForPair_BreachMaximum(uint256) (runs: 256, μ: 108416, ~: 108362)
< PairTest:testSetPlatformFeeForPair_Reset() (gas: 120551)
< PairTest:testSetSwapFeeForPair() (gas: 110894)
< PairTest:testSetSwapFeeForPair_BreachMaximum(uint256) (runs: 256, μ: 108341, ~: 108341)
< PairTest:testSetSwapFeeForPair_Reset() (gas: 126568)
---
> PairTest:testRecoverToken() (gas: 200623)
> PairTest:testSetPlatformFeeForPair() (gas: 110761)
> PairTest:testSetPlatformFeeForPair_BreachMaximum(uint256) (runs: 256, μ: 108372, ~: 108318)
> PairTest:testSetPlatformFeeForPair_Reset() (gas: 120507)
> PairTest:testSetSwapFeeForPair() (gas: 110850)
> PairTest:testSetSwapFeeForPair_BreachMaximum(uint256) (runs: 256, μ: 108297, ~: 108297)
> PairTest:testSetSwapFeeForPair_Reset() (gas: 126524)
70c70
< PairTest:testUpdateDefaultFees() (gas: 168750)
---
> PairTest:testUpdateDefaultFees() (gas: 168694)
77,79c77,79
< StablePairTest:testGetCurrentA() (gas: 32339)
---
> StablePairTest:testGetCurrentA() (gas: 32317)
81,94c81,94
< StablePairTest:testMint_NonOptimalProportion() (gas: 148783)
< StablePairTest:testMint_NonOptimalProportion_ThenBurn() (gas: 297251)
< StablePairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 247641)
< StablePairTest:testOracle_CorrectLiquidity() (gas: 285812)
< StablePairTest:testOracle_CorrectPrice() (gas: 373956)
< StablePairTest:testOracle_LiquidityAtMaximum() (gas: 277274)
< StablePairTest:testOracle_NoWriteInSameTimestamp() (gas: 205928)
< StablePairTest:testOracle_OverflowAccLiquidity() (gas: 133833)
< StablePairTest:testOracle_OverflowAccPrice() (gas: 198916)
< StablePairTest:testOracle_SimplePrices() (gas: 411162)
---
> StablePairTest:testMint_NonOptimalProportion() (gas: 148805)
> StablePairTest:testMint_NonOptimalProportion_ThenBurn() (gas: 297273)
> StablePairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 247663)
> StablePairTest:testOracle_CorrectLiquidity() (gas: 285856)
> StablePairTest:testOracle_CorrectPrice() (gas: 374022)
> StablePairTest:testOracle_LiquidityAtMaximum() (gas: 277318)
> StablePairTest:testOracle_NoWriteInSameTimestamp() (gas: 205950)
> StablePairTest:testOracle_OverflowAccLiquidity() (gas: 133899)
> StablePairTest:testOracle_OverflowAccPrice() (gas: 198982)
> StablePairTest:testOracle_SimplePrices() (gas: 411228)
96,100c96,100
< StablePairTest:testRampA() (gas: 32538)
< StablePairTest:testRampA_BreachMaxSpeed() (gas: 22808)
< StablePairTest:testRampA_BreachMaximum() (gas: 18552)
< StablePairTest:testRampA_BreachMinimum() (gas: 18488)
< StablePairTest:testRampA_MaxSpeed() (gas: 26492)
---
> StablePairTest:testRampA() (gas: 32516)
> StablePairTest:testRampA_BreachMaxSpeed() (gas: 22786)
> StablePairTest:testRampA_BreachMaximum() (gas: 18530)
> StablePairTest:testRampA_BreachMinimum() (gas: 18466)
> StablePairTest:testRampA_MaxSpeed() (gas: 26470)
102,108c102,108
< StablePairTest:testRampA_SetAtMaximum() (gas: 25248)
< StablePairTest:testRampA_SetAtMinimum() (gas: 25227)
< StablePairTest:testRampA_SwappingDuringRampingDown(uint256,uint64,uint64,uint128) (runs: 256, μ: 96046, ~: 100809)
< StablePairTest:testRampA_SwappingDuringRampingUp(uint256,uint64,uint64,uint128) (runs: 256, μ: 96277, ~: 100835)
< StablePairTest:testStopRampA() (gas: 33916)
< StablePairTest:testStopRampA_Early(uint256) (runs: 256, μ: 41806, ~: 41932)
< StablePairTest:testStopRampA_Late(uint256) (runs: 256, μ: 40653, ~: 40784)
---
> StablePairTest:testRampA_SetAtMaximum() (gas: 25226)
> StablePairTest:testRampA_SetAtMinimum() (gas: 25205)
> StablePairTest:testRampA_SwappingDuringRampingDown(uint256,uint64,uint64,uint128) (runs: 256, μ: 96029, ~: 100787)
> StablePairTest:testRampA_SwappingDuringRampingUp(uint256,uint64,uint64,uint128) (runs: 256, μ: 96283, ~: 100813)
> StablePairTest:testStopRampA() (gas: 33872)
> StablePairTest:testStopRampA_Early(uint256) (runs: 256, μ: 41763, ~: 41888)
> StablePairTest:testStopRampA_Late(uint256) (runs: 256, μ: 40610, ~: 40740)
116,119c116,119
< StablePairTest:testSwap_Token0ExactOut(uint256) (runs: 256, μ: 124546, ~: 132837)
< StablePairTest:testSwap_Token1ExactOut(uint256) (runs: 256, μ: 124681, ~: 132972)
---
> StablePairTest:testSwap_Token0ExactOut(uint256) (runs: 256, μ: 124548, ~: 132837)
> StablePairTest:testSwap_Token1ExactOut(uint256) (runs: 256, μ: 124683, ~: 132972)
121c121
< StablePairTest:testWriteObservations() (gas: 287076)
---
> StablePairTest:testWriteObservations() (gas: 287186)
124,126c124,126
< StableMathTest:testComputeLiquidityFromAdjustedBalances_ConvergeEvenWithVeryUnbalancedValues(uint256,uint256,uint256) (runs: 256, μ: 19143, ~: 13711)
< StableMathTest:testGetAmountIn(uint256,uint256,uint256) (runs: 256, μ: 24487, ~: 23987)
< StableMathTest:testGetAmountOut(uint256,uint256,uint256) (runs: 256, μ: 27088, ~: 26263)
---
> StableMathTest:testComputeLiquidityFromAdjustedBalances_ConvergeEvenWithVeryUnbalancedValues(uint256,uint256,uint256) (runs: 256, μ: 19166, ~: 13711)
> StableMathTest:testGetAmountIn(uint256,uint256,uint256) (runs: 256, μ: 24493, ~: 23987)
> StableMathTest:testGetAmountOut(uint256,uint256,uint256) (runs: 256, μ: 27107, ~: 26263)
```

1. besides these things, it seems like there's no other libs to swap out for more gas efficient ones


OliverNChalk:
Are you benchmarking deployment cost or runtime cost. Benchmarking deployment cost is a common pitfall when trying to gas golf

OliverNChalk:
Just confirmed, you seem to often _createPair in test which makes the test invalid for gas estimation purposes (please only focus on tests which have no CREATE/CREATE2 calls)

xenide:
Okay so I went through each test case and deleted those lines with have CREATE/CREATE2 operations. 

Now it seems like for `ReentrancyGuard` it's quite inconclusive. Sometimes it's more, sometimes less. 

For `owned` / `ownable`, it seems also quite inconclusive to me. Sometimes more sometimes less. 

I'm a lil lost lol, how do we proceed? The worst case scenario is that we declare that there's not enough evidence to choose one over the other and just keep the status quo?

OliverNChalk:
Did you:

1. Ensure you did not re-order the tests? (gas includes the call dispatch table so in big test files this can be quite material)
2. Isolated the measured change just to the library
3. Are benchmarking the use of onlyOwner? I suspect you are just looking at all functions which will cahnge because the function dispatch table has changed.

Please become familiar with https://github.com/ethereum/solidity/issues/4034 and understand the consequences this will have on gas bench marking. Simply renaming a function can change the gas cost of that function and all other functions in the contract

xenide:
1. I did not rename tests, rearrange tests, nor re-run any test. I simply deleted the lines corresponding to tests with contract creation in the diff in the PR description
2. okay for this, for the ownable/owned comparison, I deleted a constructor in the diff (since owned required one but ownable didn't). But I re-ran the diff again, and I'm getting the same numbers. This is expected cuz the the gas cost of constructing the object wouldn't be calculated into the test. 
3. So you're saying we should only look at tests that explicitly use the corresponding functions and ignore those that don't? Since using diff libs could potentially change the number of exposed functions and thus the binary search cost, shouldn't we look at all test gas benchmarks (less those create new contracts of cuz)

OliverNChalk:
Its not jsut about number of functions but also their names. Because each name has a bytes4 hash in the binary tree. This means changing any function name could change the gas cost of all other functions

xenide:
Yup I understand that. When the name changes it changes the traversal path (and potentially the number of splits) for the binary search. 

So when we consider whether we should switch out one library for another, should we only consider the gas usage of the direct call of the functions itself, or should we also consider the second order effects it has on other functions? After seeing the data and understanding the effects surrounding it, I'm proposing to just look at the direct use of the lib / functions themselves and not look at the second order effects. 

E.g. for `Ownable` has 3 public functions `owner` `renounceOwnership` and `transferOwnership`, while `Owned` has 2: `owner` and `setOwner`. 

This could be the explanation for why when we look at functions that directly used the modifier, like `testRampA*` which invokes factory's `onlyOwner`, the gas usage decreased across the board, while if we look at `testOracle*`, it could be that for that particular call `oracleCaller.observation()`, it took 1 more binary search compared to previously which caused the gas price to increase. In other words it is the combination of other functions in `oracleCaller` and their namings which caused us to observe this phenomenon of increased gas usage. 

So with that said, we should still go ahead with solmate's `Owned` since it's more gas efficient for execution. For `ReentrancyGuard`, there's virtually no difference in gas and in actual impl, but I propose to go with solmate's impl as it is more minimalist. Agreed?

OliverNChalk:
Agreed

xenide:
Alright, changed all instances of those lib usage to the correct ones and updated gas snapshot. Can approve if everything is good.

OliverNChalk (approved):

PR #122, opened 2022-11-22
poc: callback workflow for `swap` and `mint` 

xenide:
## Solution

- given the large difference in implementation, we might want to consider creating a new curve named something like `RebaseStablePair`
- may need another function to support flash swaps instead of rolling it into just the normal swap function
   - this is the way UniV3 does it as well  
- if the token rebases up and down, the amount of platform fee collected might be a lil strange 
- made the basic minimum case work, see `.gas-snapshot` for actual gas used
  - for mint: 
    - gas cost 106722 -> 113783 (6.61% increase)
  - for swap: 
    - gas cost 84851 -> 105116 (23.9% increase)


xenide:
closing as we're not pursuing this at this stage

PR #123, opened 2022-11-28
test: more scenarios for shares calculation 

### LINKED ISSUE #117: question: are shares still correct when we add/remove underlying tokens if there was a PnL during `adjustManagement`, but the pair has not been informed?
This is referring to AaveManager::shares, and is talking about the same pair. 

Suppose a pair has some assets managed by the AM. Some profit/loss has happened, and then the privileged AM owner decides to invest / divest some assets manually. Because `adjustManagement` does not inform the pair of gains / losses, only changes in managed amount for the current adjustment. Could this result in an inaccurate number of shares as tracked by the asset manager? I'm going to write test cases to show this 

Another test case to add: Pair1 and Pair2 both have USDC managed by the asset manager. Pair1 lets the asset manager manage some of its USDC. Then the asset manager makes some gains / losses on these assets. Then Pair2 comes along and gets some assets managed by the manager. At this point in time are the shares for Pair1 still accurate and the shares issued for Pair2 correct? We already have a test case for profit, so need to add one for the loss case

xenide:
## Motivation

- from discussion #117, I attempted to write a test case to determine if the behavior was correct. The current implementation is correct with no changes needed. Even if the pair is not synced with the latest managed figures, `adjustManagement` would still work correctly and the shares issued would be correct as well. 

## Question

- do you have any ideas on how to simulate a loss on chain? I've tried `deal` but it gives a slot not found error (same results when I give it the impl address, somehow this problem doesn't exist for USDC which is a proxy contract too) and `rewind` but that doesn't work either.


OliverNChalk:
What do you mean simulate, is this a fork test or a unit test?

For a unit test just create an asset manager mock and have it return a loss. You could also look into `mockCall()`

For a fork test, are you trying to have Aave report a loss? If so not sure how to do that but is that really important to test? It might be important for testing the AssetManager but surely its irrelevant to the pair?

So I would suggest mocking the AM or the call to the AM and producing a loss in a unit test.

test/integration/Aave.t.sol
```diff
@@ -428,6 +436,41 @@ contract AaveIntegrationTest is BaseTest
...
+        _manager.adjustManagement(
+            _pair,
+            _pair.token0() == USDC ? lAmountToManagePair : int256(0),
+            _pair.token1() == USDC ? lAmountToManagePair : int256(0)
+        );
+
+        // act - asset manager loses 10% of the assets
+        // todo: how to make manager have a loss? the following is complaining that storage slot can't be found
+        // even if I provide the impl address instead of the proxy address
+        // rewinding in time doesn't work
```
OliverNChalk: My guess is they used a packed slot, which you cannot currently find using stdStorage.

You could overwrite the Aave bytecode with a version that produces losses? E.g. find the market you are targeting and copy its source code. Then just edit the interest calc line to be negative or something. Use `vm.etch` to overwrite the bytecode, skip 30 days, and then receive the loss.

Not sure if you will encounter side effects but could be worth if you don't see a more correct way to have aave trigger a loss (check if there is anyway for a market to even report a loss?)

xenide: I looked into the docs and code. It seems that loss is not a case that is built into the model of the system. Seems like it's designed for up-only interest bearing. 

I guess forge can't find the storage slot cuz the `balanceOf` and `totalSupply` are functions with computation involved for `AToken`s instead of just simple mappings.

So I suggest, I'll just write the tests for cases to simulate profit and no cases to simulate loss. Sounds good?

OliverNChalk: Sounds good, so long as the pair is unit tested on losses (separate to Aave integration tests)

xenide: So I added a unit test case that tests for `adjustManagement` after loss. In the existing suite of unit test cases there are cases that test for syncing (through `burn`) after loss so we're good on that front

test/unit/AssetManagedPair.t.sol
```diff
@@ -141,6 +141,30 @@ contract AssetManagedPairTest is BaseTest
...
+        // act
+        _manager.adjustManagement(_pair, 20e18, 20e18);
+        lTokenAManaged = _manager.getBalance(_pair, address(_tokenA));
+
+        // assert
+        assertEq(lTokenAManaged, 20e18 + 7e18);
+        assertEq(_pair.token0Managed(), 20e18 + 10e18);
+        _pair.sync();
+        assertEq(_pair.token0Managed(), 20e18 + 7e18); // number is updated after sync
+    }
```
OliverNChalk: It would be good to add unit test cases showing that the updated number affects:

- How many shares you get for minting (i.e. you get more)
- How much underlying you get for burning (i.e. you get less)
- How the exchange for the swap changes (i.e. the asset that was lost is more expensive)

This is essentially just saying that "K" adjusts and the pair behaves normally. You can decide if you want to just test K, or write a fuzz test that proves these properties. You could theoretically have two fuzz tests, one showing that any profit results in higher K effects, one that shows any losses results in lower K effects.

OliverNChalk (approved):
Happy to merge though I think the test case is narrow/weak, could be broadened

test/unit/AssetManagedPair.t.sol
xenide: 1. for minting, the mock asset manager doesn't implement the shares system (only the `AaveManager` does) so we can't read and assert the amount of shares
2. for burning, yes we can write a test case to make sure that we indeed get less 
3. for swapping, the the swap does not get cheaper / more expensive because the pair does not sync with the asset manager on swap.

OliverNChalk: I meant:

- Produce a loss
- Call sync()
- Assert `mint` `burn` `swap` on the pair not anywhere else

xenide: Okay wrote the 3 additional test cases. For `mint` and `burn` they `sync` with the AM at the start so no need for a `sync`, but swap needs the `sync` to be updated about the loss

OliverNChalk (approved):

PR #124, opened 2022-11-28
test: differential test between our lib and balancer's

### LINKED ISSUE #70: test(LogExpMath): Differential Fuzz LogExpMath with Balancer source
To ensure that our changes (`unchecked`) did not create divergences from original behavior.

Scope:
- LogExpMath.sol
- (potentially) LogCompression.sol
- (potentially) StableOracleMath.sol

xenide:
## Solution

- while writing the tests, I realized that we could do an optimization by doing `unchecked` arithmetic on `LogCompression`. It saves quite some gas, as seen in the snapshots. And since we have shown in the fuzz tests that the behaviors are equivalent, this change should be good


OliverNChalk:
Yeah if Balancer is using a different solidity version, your best bet is to create the wrapper in their repo so you can getCode it and deploy it in our land

src/libraries/LogCompression.sol
```diff
@@ -39,18 +39,22 @@ library LogCompression {
...
      */
     function toLowResLog(uint256 value) internal pure returns (int256) {
+    unchecked {
         int256 ln = LogExpMath.ln(int256(value));
 
         // Rounding division for signed numerator
         int256 lnWithError = (ln > 0 ? ln + _HALF_LOG_COMPRESSION_FACTOR : ln - _HALF_LOG_COMPRESSION_FACTOR);
         return lnWithError / _LOG_COMPRESSION_FACTOR;
     }
+    }
```
OliverNChalk, line 50: gas optimization or causing test failures?

test/differential/LogCompression.t.sol
```diff
@@ -0,0 +1,51 @@
...
+    {
+        // we use getDeployedCode since the library contract is stateless
+        bytes memory lBytecode = vm.getDeployedCode("./reference/balancer-v2-monorepo/pkg/solidity-utils/artifacts/contracts/helpers/LogCompressionWrapper.sol/LogCompressionWrapper.json");
+        vm.etch(_balancerLogCompression, lBytecode);
+    }
+
+    function testToLowResLog(uint256 aValue) external
+    {
+        // act
+        (bool lBool, bytes memory lRes) = _balancerLogCompression.staticcall(abi.encodeWithSignature("toLowResLog(uint256)", aValue));
```
OliverNChalk: `lBool` is an example of hungarian variable names. Linus says these are only to be used by the brain damaged. As we are both smart cookies: `lBool` -> `lSuccess`

OliverNChalk (approved):
Looks really sick, I love it.

src/libraries/LogCompression.sol
xenide: optimization, as written above in my PR description... no failure....

test/differential/LogCompression.t.sol
xenide: lol okay yeah forgot about that again. Fixed

PR #129, opened 2022-12-10
test: fuzzing for `AssetManagedPair` tests

### LINKED ISSUE #127: test: fuzz tests for `AssetManagedPair`

xenide:
## Solution

- would defer merging this until this [forge issue](https://github.com/foundry-rs/foundry/issues/3848) is fixed as fuzz tests are not failing when supposed to


OliverNChalk (approved):

webthethird:
> * would defer merging this until this [forge issue](https://github.com/foundry-rs/foundry/issues/3848) is fixed as fuzz tests are not failing when supposed to

@xenide It may be a non-issue: I opened that issue, but it may just be that the fuzzer is performing automatic test case minimization. See my latest comment on that thread - I am just waiting for confirmation that this is the case before closing the issue.

PR #130, opened 2022-12-12
feat(factory): chunk initCode so it fits within contract size limit

OliverNChalk:
This addresses the first problem with our contract size, the fact we cannot store the initCode in the factory. A subsequent PR will address the size issue with StablePair


test/unit/GenericFactory.t.sol
```diff
@@ -11,6 +11,7 @@ import { GenericFactory } from "src/GenericFactory.sol";
 contract GenericFactoryTest is BaseTest {
     function testCreatePair_AllCurves(uint256 aCurveId) public {
         // assume
+        // TODO: Is this the way to test all curves?
         uint256 lCurveId = bound(aCurveId, 0, 1);
```
xenide, line 15: Yes this was the way I came up with to test creating all curves

OliverNChalk: Ahh, I forgot it was zero-indexed. Can resolve

PR #131, opened 2022-12-14
refactor: split StablePair logic into two contracts

OliverNChalk:
To enable deployment of the StablePair, it was necessary to split the logic into two contracts. The second contract contains mint/burn logic, while the former contains the rest.

This current implementation attempts to perform the minimal changes necessary. A more advanced implementation could use a fallback function to reduce the cost of function dispatch on the main contract.


src/curve/stable/StablePair.sol
```diff
@@ -356,6 +338,7 @@ contract StablePair is ReservoirPair {
         return _getCurrentAPrecise();
     }
 
+    // TODO: Is this function needed?
     function getAmountOut(address tokenIn, uint256 amountIn) public view returns (uint256 finalAmountOut) {
```
xenide, line 342: technically no since the periphery's quoter has functions to facilitate this. In this core repo we have a few tests that use it but can be replaced

src/curve/stable/StablePair.sol
```diff
@@ -367,6 +350,7 @@ contract StablePair is ReservoirPair {
         }
     }
 
+    // TODO: Do we need this function?
```
xenide, line 353: only for viewing / reporting purposes to track the growth of the value of the LP token over time

src/GenericFactory.sol
```diff
@@ -153,4 +157,24 @@ contract GenericFactory is IGenericFactory, Owned {
...
+
+    event Deployed(address aAddress, bytes32 aSalt);
+
+    /// @notice Deploys a given contract and salt using CREATE2.
+    /// @param  aInitCode   The contract init code to deploy.
+    /// @param  aSalt       A user-provided value that can be used to produce
+    ///                     varying addresses for the same init code.
+    /// @return rContract The address of the newly deployed contract.
+    function deploy(bytes memory aInitCode, bytes32 aSalt) external payable returns (address rContract) {
+        require(_deployInProgress, "AUTH: ONLY CHILDREN CAN CALL");
```
xenide: should we change the error msg to `FACTORY: ONLY_CHILDREN_CAN_CALL` to be more in line with the existing style?

src/curve/stable/StablePair.sol
```diff
@@ -113,73 +123,45 @@ contract StablePair is ReservoirPair {
...
-        lastInvariant = uint192(newLiq);
-        lastInvariantAmp = _getCurrentAPrecise();
-
-        emit Mint(msg.sender, amount0, amount1);
-
-        _managerCallback();
     }
 
+    // TODO: Test re-entrancy.
+    // TODO: Should we use fallback?
```
xenide, line 148: as in `fallback() external {}` ? Is the motivation gas savings? Even if we use it, we can't use it for both `mint` and `burn`, only one at most right?

src/curve/stable/StableMintBurn.sol
```diff
@@ -0,0 +1,248 @@
...
+        lastInvariantAmp = _getCurrentAPrecise();
+
+        emit Burn(msg.sender, amount0, amount1);
+
+        _managerCallback();
+    }
+
+    /// @inheritdoc IPair
+    function swap(int256, bool, address, bytes calldata) external pure returns (uint256) {
+        revert();
```
xenide:
```suggestion
        revert("SMB: NOT_IMPLEMENTED");
```

src/curve/stable/StablePair.sol
OliverNChalk: We can use it for both, any call that does not match a valid function will forward to StableMintBurn. So this means both `mint` and `burn` will forward

src/curve/stable/StableMintBurn.sol
OliverNChalk: Sounds good, will do

src/GenericFactory.sol
OliverNChalk: Yes, will do

PR #135, opened 2022-12-15
refactor: reduce StablePair initCode size

OliverNChalk:


xenide (approved):
I guess the codecov thing is just acting up? Nothing to be concerned about right?

PR #136, opened 2022-12-16
test: flash swap

### LINKED ISSUE #81: test: flash swap

xenide:
## Solution

- decided to modify the signature of the reservoirCallee function, so that the callee knows how much to pay for the swap. Although in UniV2 their signature is also only amount0/1Out and no amount in


test/unit/FlashSwap.t.sol
```diff
@@ -0,0 +1,55 @@
...
+            vm.revertTo(lBefore);
+        }
+    }
+
+    function setUp() public {
+        _pairs.push(_constantProductPair);
+        _pairs.push(_stablePair);
+    }
+
+    function reservoirCall(address aSender, int256 aAmount0, int256 aAmount1, bytes calldata aData) external {
```
OliverNChalk, line 27: lints

test/unit/FlashSwap.t.sol
```diff
@@ -0,0 +1,55 @@
...
+    function testSwap_FlashSwap_ExactOut(uint256 aSwapAmt) external allPairs {
+        // assume
+        int256 lSwapAmt = -int256(bound(aSwapAmt, 1, INITIAL_MINT_AMOUNT / 2));
+
+        // act
+        _pair.swap(lSwapAmt, false, address(this), "some bytes");
+
+        // assert
+        assertEq(_tokenB.balanceOf(address(this)), uint256(-lSwapAmt));
+    }
```
OliverNChalk, line 59: Want to add negative tests that show the swap fails if you don't pay?

xenide: okay added

test/unit/FlashSwap.t.sol
xenide: fixed

PR #138, opened 2022-12-20
wip: non-trivial gas savings by packing index into reserves slot

OliverNChalk:
Question: Should we leave the reserves alone and just store the index in the same slot as the `bool lock`  for re-entrancy guard? You could also turn this on its head and ask can we get away wit ha 24bit timestamp (though this seems much dodgier unless you `/10` each timestamp).


xenide (approved):
Reviewed the failing test, can confirm that it is due to the precision change. You can increase the tolerance for the test to pass.

src/asset-management/AaveManager.sol
```diff
@@ -51,18 +51,18 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
-    function _getBalance(IAssetManagedPair aOwner, address aToken) private view returns (uint112 rTokenBalance) {
+    function _getBalance(IAssetManagedPair aOwner, address aToken) private view returns (uint104 rTokenBalance) {
         address lAaveToken = _getATokenAddress(aToken);
         uint256 lTotalShares = totalShares[lAaveToken];
         if (lTotalShares == 0) {
             return 0;
         }
         rTokenBalance =
-            uint112(shares[aOwner][aToken] * IERC20(lAaveToken).balanceOf(address(this)) / totalShares[lAaveToken]);
+            uint104(shares[aOwner][aToken] * IERC20(lAaveToken).balanceOf(address(this)) / totalShares[lAaveToken]);
```
OliverNChalk, line 65: Overflow?

xenide: I think we should be good cuz `balanceOf` returns a uint256 which the `uint104` will be implicitly cast into during the operation

OliverNChalk: There is no overflow checking when you cast uint256 to uint104 in solidity pretty sure?

xenide: that's correct. No overflow check. But with the amounts we are managing, it's unlikely that we'd reach that limit. However, if we're still worried, we can use openzeppelin's `SafeCast::toUint104()` to ensure that it is indeed within range

OliverNChalk: Yeah, I'd probably prefer a safeCast here even though its extremely unlikely its still possible

xenide: okay created #140 to make sure we don't miss it. 

Resolving for now

PR #139, opened 2022-12-21
perf: pack lock

OliverNChalk:
This PR packs the lock storage variable into Slot0 by reducing `blockTimestampLast` from uint32 to uint31 (it has a getter and setter to support this non-standard packing).


test/__fixtures/BaseTest.sol
```diff
@@ -122,6 +122,7 @@ abstract contract BaseTest is Test {
         int56 aLiq,
         uint32 aTime
     ) internal {
+        require(aTime < 2 ** 31, "TIMESTAMP TOO BIG");
```
xenide, line 130: technically this can be `aTime <= 2 ** 31` right?

xenide (commented):
only spotted one thing, the rest looks sick

test/__fixtures/BaseTest.sol
OliverNChalk: Hmm, I think not right, because the max value for uint31 is 2**31 - 1 right? (it's always -1 because of 0)? That said I was lucky, didnt think too hard about this

xenide: I think you're right, the max value should be 2 ** 31 - 1, so your require statement is fine

xenide (approved):

PR #141, opened 2022-12-22
perf: minimize SLOAD/SSTORE

OliverNChalk:
Notable changes:
- Skim now triggers an oracle update & writes blockTimestampLast

Tests needed (run the tests against before/after contracts, I can do it if necessary as it will require a little git trickery): 
- [x] `lastInvariant` doesn't seem to be tested in `burn`. I broke it and no tests failed.
- [x] Test that asserts whether or not we want oracle update reserves to include AssetManager PNL. The tests currently pass regardless of whether you include AssetManager PNL or not.
- [x] A test that checks StablePair reverts with overflow if you try to increase the reserves beyond uint104


test/unit/ConstantProductPair.t.sol
```diff
@@ -681,6 +682,7 @@ contract ConstantProductPairTest is BaseTest, IReservoirCallee {
 
     function testPlatformFee_DisableReenable() external {
         // sanity
+        assertGt(_constantProductPair.platformFee(), 100);
```
xenide, line 685: why specifically asserting that it is greater than 100?

xenide: why specifically asserting that it is greater than 100?

src/ReservoirPair.sol
```diff
@@ -19,7 +18,7 @@ abstract contract ReservoirPair is AssetManagedPair, OracleWriter {
```
xenide: lint, can remove the local variable

test/unit/ConstantProductPair.t.sol
```diff
@@ -642,6 +642,7 @@ contract ConstantProductPairTest is BaseTest, IReservoirCallee {
 
     function testPlatformFee_Disable() external {
         // sanity
+        assertGt(_constantProductPair.platformFee(), 100);
```
xenide: same here

src/ReservoirPair.sol
```diff
@@ -44,9 +43,9 @@ abstract contract ReservoirPair is AssetManagedPair, OracleWriter {
     }
 
     // update reserves and, on the first call per block, price and liq accumulators
-    function _update(uint256 aBalance0, uint256 aBalance1, uint104 aReserve0, uint104 aReserve1) internal override {
+    function _updateAndUnlock(uint256 aBalance0, uint256 aBalance1, uint104 aReserve0, uint104 aReserve1) internal {
         // TODO: Cache this load?
-        (uint32 lBlockTimestampLast, bool lLocked) = _splitSlot0Timestamp(_slot0.packedTimestamp);
+        (uint32 lBlockTimestampLast,) = _splitSlot0Timestamp(_slot0.packedTimestamp);
         require(aBalance0 <= type(uint104).max && aBalance1 <= type(uint104).max, "CP: OVERFLOW");
```
xenide:
```suggestion
        require(aBalance0 <= type(uint104).max && aBalance1 <= type(uint104).max, "RP: OVERFLOW");
```

test/unit/ConstantProductPair.t.sol
OliverNChalk: That it's non-zero, could change to >0 if we want. Did 100 to avoid chances of roudning platform mint to zero, though I think it wont even if platformFee is 1

src/curve/stable/StableMintBurn.sol
```diff
@@ -165,11 +162,11 @@ contract StableMintBurn is ReservoirPair {
     /// @dev Originally
     /// https://github.com/saddle-finance/saddle-contract/blob/0b76f7fb519e34b878aa1d58cffc8d8dc0572c12/contracts/SwapUtils.sol#L319.
     /// @return liquidity The invariant, at the precision of the pool.
-    function _computeLiquidity(uint256 lReserve0, uint256 lReserve1) internal view returns (uint256 liquidity) {
+    function _computeLiquidity(uint256 aReserve0, uint256 aReserve1) internal view returns (uint256 liquidity) {
```
xenide: if we want to standardize the naming throughout, do we want to use `rLiquidity` instead?

test/unit/ConstantProductPair.t.sol
```diff
@@ -737,4 +738,25 @@ contract ConstantProductPairTest is BaseTest, IReservoirCallee {
         uint256 lNewShares = _constantProductPair.balanceOf(address(_platformFeeTo)) - lPlatformShares;
         assertLt(lNewShares, lPlatformShares);
     }
+
+    function testSync() external {
```
xenide: we also have a `testSync` in `AssetManagedPair.t.sol` but that's specifically testing some asset management things. For this particular test, I suggest moving it to `ReservoirPair.t.sol` as the `sync` function is defined in `ReservoirPair`

xenide (commented):
I considered the change in the behavior of `skim` again. Some thoughts: 

1. the purpose of `skim` is to force actual balances to match reserves. So technically it should not re-write reserves since that should not change. Only the balance should change.    
   - Do you think creating and using another internal function `_unlock` would be appropriate here? 


Btw do you need me to write those additional tests listed above?

src/curve/stable/StableMintBurn.sol
OliverNChalk: Yes, sounds good

test/unit/ConstantProductPair.t.sol
OliverNChalk: Sounds good, then we should extend the test to work on both

OliverNChalk:
If you're able to implement your comments & then expand the tests, I think that's the next step for closing this. I'm going to have to take a break from pushing changes in for a bit. We should have the two swap function fairly golfed by now.

I'll probably do another sweep right before audit but I think you're good to just open PRs/commit directly to dev while you address `TODO` and `PERF` comments in the codebase.

xenide:
> * [ ]  `lastInvariant` doesn't seem to be tested in `burn`. I broke it and no tests failed.

What was the change that broke `lastInvariant`? Side note: currently they are private variables, and there doesn't seem to be a reason to make them public is there? Balancer also did not expose them as public variables 

> * [ ]  Test that asserts whether or not we want oracle update reserves to include AssetManager PNL. The tests currently pass regardless of whether you include AssetManager PNL or not.

I'm trying to make sure I understand this correctly. Are you referring to, whether to let `_updateOracle` be informed about the asset manager's PnL? When we discussed this previously, we agreed that only `mint`/`burn` needs to sync with the asset manager, while `swap` doesn't need to.

OliverNChalk:
> What was the change that broke lastInvariant? Side note: currently they are private variables, and there doesn't seem to be a reason to make them public is there? Balancer also did not expose them as public variables.

Writing reserves before/after asset manager sync. I'd use tests to make sure we are always either writing reserves before OR after we sync with asset manager. Also make sure we are writing reserves before the changes from mint/burn/swap are applied

OliverNChalk:
>I'm trying to make sure I understand this correctly. Are you referring to, whether to let _updateOracle be informed about the asset manager's PnL? When we discussed this previously, we agreed that only mint/burn needs to sync with the asset manager, while swap doesn't need to.

Mint and burn write the oracle so no need to worry about swap

xenide:
@OliverNChalk I have implemented all the required tests and made `skim` not update the oracle by just unlocking. You can approve the PR if all is good, thanks.

Not sure why the coverage is failing? Could it be the key in `Constants.sol`?

OliverNChalk:
Yes look at my commits where it doesn't fail and you'll see what's up. You need to first run `npm run coverage` locally to get it to generate the new key for you (this will actually place Constants.sol in a broken state so just `git checkout src/library/Constants.sol` to get it back to a known value [we use a bash script to swap between coverage mode and optimized mode]).

src/curve/stable/StableMintBurn.sol
```diff
@@ -137,40 +135,39 @@ contract StableMintBurn is ReservoirPair {
...
-
-        lastInvariant = uint192(_computeLiquidity(_slot0.reserve0, _slot0.reserve1));
+        uint256 lBalance0 = _totalToken0();
+        uint256 lBalance1 = _totalToken1();
+        lastInvariant = uint192(_computeLiquidity(lBalance0, lBalance1));
         lastInvariantAmp = _getCurrentAPrecise();
-
         emit Burn(msg.sender, amount0, amount1);
 
+        _updateAndUnlock(lBalance0, lBalance1, uint104(lReserve0), uint104(lReserve1), lBlockTimestampLast);
```
OliverNChalk: Are these `uint104` casts extraneous/no-ops?

OliverNChalk: Can merge after this is resolved

xenide: yes they seem superfluous, and the compiler actually makes it a no-op. 

Fixed

PR #144, opened 2022-12-26
bug: last invariant not tested

xenide:
## Solution

- created the test case to show that if we used reserves instead of balances, no platformFee would be minted and the test will fail. 
- if all is good, I will revert to the correct impl and merge into the `perf/minimize-sload-store` branch. Or is there other "git trickery" things you wanna do?


OliverNChalk:
Looks good, two options to close:

1. Close PR and just cherry pick test change commit (might still break the hash)
2. Squash merge PR & edit squash to remove the bug

xenide:
> Looks good, two options to close:
> 
> 1. Close PR and just cherry pick test change commit (might still break the hash)
> 2. Squash merge PR & edit squash to remove the bug

How about a 3rd option where I push a commit to remove the bug and then we squash merge?

OliverNChalk:
Sounds good, get CI passing then squash

PR #145, opened 2022-12-27
test: oracle update include AM PnL

xenide:
## Solution

- because we cannot test solely oracle writing, we have to do it through `mint`, `burn`, and `sync`


OliverNChalk:
The tests pass when I checkout the branch? only testSyncManaged fails?

xenide:
I think it's some bug of forge's. The assertions fail but the tests themselves don't fail. 
Screenshot 2022-12-27 at 11 52 27 AM

OliverNChalk:
Oh weird, I got that too at one point. Was like when I had an assertion fail in a callback to the test contract

PR #148, opened 2022-12-28
test: more test cases for coverage 

### LINKED ISSUE #134: test: missing test cases/coverage
Here are some missing test scenarios that should be added:

- [x] Disabling platform fee on `ConstantProductPair` sets `kLast` to zero. This would be tested by creating a pair, accumulating some fee, setting platform fee to zero, claiming the fee for the pair, then ensuring from that point onwards no more fees are charged until the fee is re-enabled.
- [x] Testing `IReservoirCallee` callback in `ConstantProductPair`. Could be covered by the flashSwap test case?
- [x] Testing `IReservoirCallee` callback in `StablePair`. Could be covered by the flashSwap test case?
- [ ] Testing StablePair getAmountOut with invalid token (or remove this from StablePair).
- [ ] Testing StablePair mint/burn while ramping
- [ ] Testing AaveManager set{Lower,Upper}Threshold <- already exist
- [ ] Testing `_checkedTransfer` underlying ERC-20 transfer fails triggering full revert
- [ ] Test swap min int256 triggering a cast error
- [ ] Adjusting platformFee affects pending platform shares. E.g. platform fee could be adjusted to 100% to claim all growth in K since last mint/burn. Is this a problem?

xenide:
## Motivation

> * [ ]  Testing StablePair getAmountOut with invalid token (or remove this from StablePair).

   - removed `getAmountOut` and refactored tests 

> * [ ]  Testing StablePair mint/burn while ramping

   - implemented 

> * [ ]  Testing AaveManager set{Lower,Upper}Threshold <- already exist

   - already exists as `testSet{Lower,Upper}Threshold_*`

> * [ ]  Testing `_checkedTransfer` underlying ERC-20 transfer fails triggering full revert

   - implemented 

> * [ ]  Test swap min int256 triggering a cast error

   - implemented - see note below 

> * [ ]  Adjusting platformFee affects pending platform shares. E.g. platform fee could be adjusted to 100% to claim all growth in K since last mint/burn. Is this a problem?

   - made an issue for this 
 
> * [ ]  platformFee minting, my 2 new tests caught a bug here that would've gone undetected a week ago

- you referring to `testPlatformFee_Disable` and `testPlatformFee_DisableReenable`? 
- anyway, I discovered a "bug", or at least a difference in implementation between our two curves. The StablePair assumes that platform fee is always on (i.e. never 0), and doesn't reset the counter when we turn it off. So I've changed the impl to match ConstantProductPair, which sets the `lastInvariant` to 0 if platform fee is off  

> * [ ]  oracle _update semantics (should it write the new reserves or previous reserves). The answer is previous but there are limited tests enforcing this, so a bug was introduced that I am now testing/fixing

- implemented

## Solution

- we don't seem to have a check for the swap amount being min int256, it would just fail from an arithmetic error and not a cast error 
   - we should add a check to make sure that the swap amount is not `type(int256).min`?
   - the `SafeCast::toUint256()` doesn't solve this problem cuz it requires the `int256` to be positive in the first place


test/unit/ConstantProductPair.t.sol
```diff
@@ -659,6 +668,7 @@ contract ConstantProductPairTest is BaseTest, IReservoirCallee {
         assertGt(lToken0.balanceOf(address(_constantProductPair)), INITIAL_MINT_AMOUNT);
         assertEq(lToken1.balanceOf(address(_constantProductPair)), INITIAL_MINT_AMOUNT);
         assertEq(_constantProductPair.platformFee(), DEFAULT_PLATFORM_FEE);
+        // why are we getting the balance of _factory?
```
xenide: what's the reason for this?

OliverNChalk:
> you referring to testPlatformFee_Disable and testPlatformFee_DisableReenable?

Yep

> we should add a check to make sure that the swap amount is not type(int256).min?

If there is no risk of the current code then we do not need to change it. We just need to make sure the code behaves correctly/safely when min int256 is used

test/unit/ConstantProductPair.t.sol
OliverNChalk: Clearly a bug, should be _platformFeeTo

test/unit/ConstantProductPair.t.sol
```diff
@@ -726,6 +735,7 @@ contract ConstantProductPairTest is BaseTest, IReservoirCallee {
         // act - swap once at half volume, again with platform fee.
         vm.prank(address(_factory));
         _constantProductPair.setCustomPlatformFee(type(uint256).max);
+        _constantProductPair.burn(address(this));
```
xenide, line 738: fyi this we need this statement to set the `kLast = 0`, so that the next burn actually mints the platform fee. Without this, the next `burn` would set `kLast = 0`, resulting in no new shares minted, which would make the `assertLt` trivially true

xenide:
> If there is no risk of the current code then we do not need to change it. We just need to make sure the code behaves correctly/safely when min int256 is used

Currently, when `type(int256).min` is passed in as an argument, the casting into `uint256` would throw an arithmetic overflow, cuz the int256.max is 1 smaller than int256.min in terms of magnitude. So this behavior is correct and is something we expect and want. Therefore I think no change is needed. 

Then I also thought about other cases, albeit admittedly they're kinda irrelevant in this case: 
1. for larger enough numbers, it would overflow during the arithmetic operations (* and +) in `_getAmountIn` or `_getAmountOut` 
2. even if the above doesn't happen, if the resulting amount is greater than `uint104.max`, it would fail at the `_updateAndUnlock` stage cuz the number exceeds `uint104`

test/unit/ConstantProductPair.t.sol
OliverNChalk: Sounds good

src/curve/stable/StableMintBurn.sol
```diff
@@ -93,7 +93,7 @@ contract StableMintBurn is ReservoirPair {
         lReserve0 += uint104(lFee0);
         lReserve1 += uint104(lFee1);
 
-        (uint256 _totalSupply, uint256 oldLiq) = _mintFee(lReserve0, lReserve1);
+        (bool feeOn, uint256 _totalSupply, uint256 oldLiq) = _mintFee(lReserve0, lReserve1);
```
OliverNChalk: variable convention

src/curve/stable/StableMintBurn.sol
```diff
@@ -128,7 +130,7 @@ contract StableMintBurn is ReservoirPair {
 
         uint256 liquidity = balanceOf[address(this)];
 
-        (uint256 _totalSupply,) = _mintFee(lReserve0, lReserve1);
+        (bool feeOn, uint256 _totalSupply,) = _mintFee(lReserve0, lReserve1);
```
OliverNChalk: Variable convention

src/curve/stable/StableMintBurn.sol
```diff
@@ -167,31 +171,40 @@ contract StableMintBurn is ReservoirPair {
...
+                StableMath._computeLiquidityFromAdjustedBalances(lAdjustedReserve0, lAdjustedReserve1, _getNA());
         }
     }
 
-    function _mintFee(uint256 lReserve0, uint256 lReserve1) internal returns (uint256 _totalSupply, uint256 d) {
+    function _mintFee(uint256 lReserve0, uint256 lReserve1)
+        internal
+        returns (bool feeOn, uint256 _totalSupply, uint256 d)
+    {
+        feeOn = platformFee > 0;
```
OliverNChalk: Variable convention

src/curve/stable/StableMintBurn.sol
```diff
@@ -167,31 +171,40 @@ contract StableMintBurn is ReservoirPair {
...
-                uint256 liquidity = numerator / denominator;
-
-                if (liquidity != 0) {
-                    address platformFeeTo = factory.read(PLATFORM_FEE_TO_NAME).toAddress();
-
-                    _mint(platformFeeTo, liquidity);
-                    _totalSupply += liquidity;
+        if (feeOn) {
+            uint256 _dLast = lastInvariant;
+            if (_dLast != 0) {
```
OliverNChalk: Variable convetion

test/unit/ReservoirPair.t.sol
```diff
@@ -119,8 +119,22 @@ contract ReservoirPairTest is BaseTest {
...
 
         // assert
-        (, , , uint16 lIndex) = _pair.getReserves();
+        (,,, uint16 lIndex) = _pair.getReserves();
         Observation memory lObs = _oracleCaller.observation(_pair, lIndex);
-        assertEq(lObs.logAccLiquidity, 470050);
+        assertEq(lObs.logAccLiquidity, 470_050);
+    }
+
+    function testCheckedTransfer_RevertWhenTransferFail() external allPairs {
```
OliverNChalk, line 127: Should test two types of fails (does mockCall let you trigger a revert?):

- Return false
- Revert

src/curve/stable/StableMintBurn.sol
```diff
@@ -140,8 +142,10 @@ contract StableMintBurn is ReservoirPair {
 
         uint256 lBalance0 = _totalToken0();
         uint256 lBalance1 = _totalToken1();
-        lastInvariant = uint192(_computeLiquidity(lBalance0, lBalance1));
-        lastInvariantAmp = _getCurrentAPrecise();
+        if (feeOn) {
+            lastInvariant = uint192(_computeLiquidity(lBalance0, lBalance1));
+            lastInvariantAmp = _getCurrentAPrecise();
+        }
```
OliverNChalk: We write lastInvariant here as well as in `_mintFee` which seems like pretty confusing/ugly code organization. On a slightly side note, what is the reason we can't just always write the last invariant?

OliverNChalk (changes requested):

test/unit/ReservoirPair.t.sol
xenide: unfortunately the current forge-std cheatcodes do not include mocking a revert (although this is something I've found myself and others desiring multiple times). 

So if we can't mock a call, the next best thing is to manually set the wallet balance to 0 and make it revert. Added the test case

src/curve/stable/StableMintBurn.sol
xenide: This approach is following what we do for `ConstantProductPair`. Uni v2 has this check before writing `kLast`. I posit that they do this to save 1 `SSTORE` if it ain't necessary.

Alright so, after thinking about this for quite a bit, it seems to me that always writing `lastInvariant` at the end of `mint`/`burn` is not a problem. But the `if (_dLast != 0)` check and `else if (lastInvariant != 0) lastInvariant = 0;` set is essential to cater for the case of turning the platform fee on and off.  

Given that we expect all our pairs to have some form of platform fee from launch, I propose we get rid of this `feeOn` check for both curves.

src/curve/stable/StableMintBurn.sol
xenide: fixed

src/curve/stable/StableMintBurn.sol
OliverNChalk: Sounds good, i like the simplicity

xenide:
while refactoring for the issue above, I discovered that if our platformFee is off, `rD` will not be calculated and that I wrote a test case to show that it fails (`testMint_PlatformFeeOff`). 

So I've fixed the behavior by always calculating `rD` no matter if platformFee is on or off.

src/curve/stable/StableMintBurn.sol
xenide: fixed

src/curve/stable/StableMintBurn.sol
```diff
@@ -167,31 +167,40 @@ contract StableMintBurn is ReservoirPair {
...
-                uint256 numerator = _totalSupply * (d - _dLast) * _platformFee;
-                uint256 denominator = (FEE_ACCURACY - _platformFee) * d + _platformFee * _dLast;
-                uint256 liquidity = numerator / denominator;
-
-                if (liquidity != 0) {
-                    address platformFeeTo = factory.read(PLATFORM_FEE_TO_NAME).toAddress();
-
-                    _mint(platformFeeTo, liquidity);
-                    _totalSupply += liquidity;
+    function _mintFee(uint256 lReserve0, uint256 lReserve1)
```
OliverNChalk: variable convention

src/curve/stable/StableMintBurn.sol
```diff
@@ -167,31 +167,40 @@ contract StableMintBurn is ReservoirPair {
...
+        );
+        if (lFeeOn) {
+            uint256 lDLast = lastInvariant;
+            if (lDLast != 0) {
+                if (rD > lDLast) {
+                    // @dev `platformFee` % of increase in liquidity.
+                    uint256 lPlatformFee = platformFee;
+                    uint256 lNumerator = rTotalSupply * (rD - lDLast) * lPlatformFee;
+                    uint256 lDenominator = (FEE_ACCURACY - lPlatformFee) * rD + lPlatformFee * lDLast;
+                    uint256 lLiq = lNumerator / lDenominator;
```
OliverNChalk:
```suggestion
                    uint256 lPlatformShares = lNumerator / lDenominator;
```

src/curve/stable/StableMintBurn.sol
```diff
@@ -167,31 +167,40 @@ contract StableMintBurn is ReservoirPair {
...
+
+                    if (lLiq != 0) {
+                        address platformFeeTo = factory.read(PLATFORM_FEE_TO_NAME).toAddress();
```
OliverNChalk:
```suggestion
                        address lPlatformFeeTo = factory.read(PLATFORM_FEE_TO_NAME).toAddress();
```

OliverNChalk (approved):

src/curve/stable/StableMintBurn.sol
xenide: fixed

src/curve/stable/StableMintBurn.sol
xenide: fixed

src/curve/stable/StableMintBurn.sol
xenide: fixed

PR #150, opened 2023-01-04
build: using via-ir

### LINKED ISSUE #137: build: investigate using `via-ir`
`via-ir` may produce a more optimized contract, though there is risk of optimizer bug(s) that we should be aware of

xenide:
## Solution

- compilation seems to take many times longer with this option enabled 
- gas usage seems to have decreased, on average
- correctness seems to have been maintained, as all tests still pass, except for that one commented below 
- note: upgraded `forge-std` version from v1.0.0 to v1.2.0 as there were [complaints](https://github.com/foundry-rs/forge-std/issues/207) that v1.0.0 was making things compilations `via-ir` slower. After the upgrade it was a lot faster (232s vs 700s)
- Can't do anything about the linting fail for now unless we use comments `/// @solidity memory-safe-assembly`, but this way of annotating will be phased out soon


OliverNChalk:
Interesting that solc will refuse to optimize unless the whole codebase/call-chain is memory safe (makes sense for the call-chain, less so whole codebase). Can you identify what block is in the call chain, maybe we can fork and annotate the dependency?

xenide:
so turns out, there was no need to manually insert the `memory-safe` annotations for the libraries. I just had to upgrade `forge-std` and `solmate` to the latest version and they seem to have gotten with the trend. 

Also it seems that we don't need the `memory-safe` annotations for our assembly for it to compile properly, but somehow the annotations are needed for the libs for it to compile. Not sure why the annotation is not needed for our own code

test/unit/ConstantProductPair.t.sol
```diff
@@ -305,8 +305,7 @@ contract ConstantProductPairTest is BaseTest, IReservoirCallee {
 
         // act
         for (uint256 i = 0; i < lMaxObservations + 4; ++i) {
-            vm.roll(block.number + 1);
-            vm.warp(block.timestamp + 5);
+            _stepTime(5);
```
xenide, line 308: this is a weird bug that popped up after switching to `via-ir`. To me the two implementations are 100% the same, but somehow the previous one gave an assertion failure in the only assertion of the test, when built with `via-ir`

OliverNChalk: Nice that you isolated to a single commit (16d043e201853d0724d3d9c4fdc23bcab57c9ee5), this could be a bug in `solc`, there have been quite a few optimizer bugs... which means we should only use via-ir if gas saving is fairly significant

OliverNChalk: We should consider reporting this bug to `solc`. We could also diff the assembly produced

OliverNChalk:
> so turns out, there was no need to manually insert the `memory-safe` annotations for the libraries. I just had to upgrade `forge-std` and `solmate` to the latest version and they seem to have gotten with the trend.
> 
> Also it seems that we don't need the `memory-safe` annotations for our assembly for it to compile properly, but somehow the annotations are needed for the libs for it to compile. Not sure why the annotation is not needed for our own code

When you annotate memory-safe, solc can move stack values into memory to avoid stack too deep. So if there are libs missing the annotations, solc might not be able to avoid stack too deep errors - hence the compilation issues.

src/GenericFactory.sol
```diff
@@ -85,6 +92,7 @@ contract GenericFactory is IGenericFactory, Owned {
             mstore(lInitCode, add(lByteCodeLength, 0x40))
             mstore(0x40, add(lFreeMem, 0x40))
         }
+        // NB: Releasing back to solidity after the loop is not memory safe.
```
xenide: Moved from the comment above. Not sure if this is still relevant? We're just returning `lInitCode` after this assembly block so it should be fine? 

I guess it's also possible to do a `return` within the assembly as well?

test/unit/ConstantProductPair.t.sol
xenide: I've tried `forge debug` that specific test but somehow the debugger just doesn't even run, so I'm trying to get the compiler to do output some asm now

OliverNChalk: should be an `--extra-output-file evm.asm` option you can use. then diff the output

src/GenericFactory.sol
```diff
@@ -63,8 +68,10 @@ contract GenericFactory is IGenericFactory, Owned {
             address lPointer = lByteCode[i];
             uint256 lSize = lPointer.code.length - 0x01;
 
-            // TODO: Go check/annotate all asm for memory safety.
-            assembly {
+            // SAFETY:
+            // This assembly block is memory safe because it writes to a region already
+            // allocated via the previous assembly block
+            assembly ("memory-safe") {
```
OliverNChalk: When we release back to solidity between assembly blocks, the memory model must be correct. We clearly copy `lSize` into memory and then release back to solidity, which is not safe. We need to write this entire loop in assembly, or write `0x40` slot in between iterations. Given gas efficiency is not a huge worry, simplest fix would be to update the free memory pointer before returning back to solidity.

src/GenericFactory.sol
OliverNChalk: The loop is in solidity with asm in the middle that writes memory without updating the pointer. So the return here is not the problem but the unsound loop is the problem

src/GenericFactory.sol
```diff
@@ -147,6 +155,9 @@ contract GenericFactory is IGenericFactory, Owned {
         // TODO: Test that _loadCurve errors for invalid indexes.
         bytes memory lInitCode = _loadCurve(aCurveId, lToken0, lToken1);
 
+        // SAFETY:
+        // This assembly block is memory safe as no previous memory is overwritten and
+        // the free memory pointer is still correct
```
OliverNChalk: Statement is weird, it should just be "this assembly block does not write to memory"

src/GenericFactory.sol
```diff
@@ -193,7 +204,11 @@ contract GenericFactory is IGenericFactory, Owned {
         require(_deployInProgress, "FACTORY: ONLY_CHILDREN_CAN_CALL");
 
         bytes memory lInitCode = getBytecode(aCodeKey, aToken0, aToken1);
-        assembly {
+
+        // SAFETY:
+        // This assembly block is memory safe as no previous memory is overwritten and
+        // the free memory pointer is still correct
```
OliverNChalk: No memory is written

src/curve/stable/StablePair.sol
```diff
@@ -109,7 +109,11 @@ contract StablePair is ReservoirPair {
     function mint(address) external override returns (uint256) {
         // DELEGATE TO StableMintBurn
         address lTarget = MINT_BURN_LOGIC;
-        assembly {
+
+        // SAFETY:
+        // This assembly block is memory safe as it does not manipulate the free memory pointer
+        // and that the calldata and return value each do not exceed 64 bytes
```
OliverNChalk: Why does calldata/returndata not exceed 64 bytes, it could easily exceed 64 bytes right?

This block is memory safe because it does not return back to solidity (see the in asm revert and return)

src/curve/stable/StablePair.sol
```diff
@@ -127,7 +131,11 @@ contract StablePair is ReservoirPair {
     function burn(address) external override returns (uint256, uint256) {
         // DELEGATE TO StableMintBurn
         address lTarget = MINT_BURN_LOGIC;
-        assembly {
+
+        // SAFETY:
+        // This assembly block is memory safe as it does not manipulate the free memory pointer
+        // and that the calldata and return value each do not exceed 64 bytes
```
OliverNChalk: Same as above

OliverNChalk (changes requested):

src/GenericFactory.sol
xenide: I see.. so since we have fixed the unsound loop [here](https://github.com/vexchange/v3-core/pull/150#discussion_r1081022846), this comment is no longer relevant, have deleted it

src/GenericFactory.sol
xenide: okay so I fixed it by updating the free memory pointer after `extcodecopy`, and also updated the docs to say that the free memory pointer has been updated, thus it's safe.

src/GenericFactory.sol
xenide: ah okay. 

I put the phrasing as the following to accentuate the fact that this block IS memory safe. 
```
// This assembly block is memory safe as it does not write to memory
```

src/GenericFactory.sol
xenide: okay fixed like the prev comment

src/curve/stable/StablePair.sol
xenide: I see. So whenever we return/revert from the assembly with a wonky free memory pointer, it is still memory safe? Cuz after the return, the caller function would have its own memory space / context and that's not been affected? 

For discussion purposes, in this case of `mint/burn` `delegatecall`, both functions' args are just address (within 32 bytes) and the return data are either 32 bytes for mint or 64 bytes for burn. I notice that both don't exceed the 64 bytes of "scratch space" (from 0x00 - 0x3f), that's why I made the comment

src/curve/stable/StablePair.sol
xenide: fixed as above

src/GenericFactory.sol
```diff
@@ -53,7 +53,12 @@ contract GenericFactory is IGenericFactory, Owned {
 
         bytes memory lInitCode;
         uint256 lFreeMem;
-        assembly {
+
+        // SAFETY:
+        // This assembly block is memory safe because it simply loads the next free memory slot,
+        // gets the next free memory slot by adding the correct number of bytes (0x20),
+        // and does not overwrite previous memory
```
OliverNChalk: "This assembly block is memory safe because" is implied and does not need to be re-written every time. I propose:

```solidity
// SAFETY:
Writes 1 word of new memory, updates the pointer before returning
```

src/curve/stable/StablePair.sol
```diff
@@ -109,7 +109,10 @@ contract StablePair is ReservoirPair {
     function mint(address) external override returns (uint256) {
         // DELEGATE TO StableMintBurn
         address lTarget = MINT_BURN_LOGIC;
-        assembly {
+
+        // SAFETY:
+        // This assembly block is memory safe as it does not return to solidity
```
OliverNChalk: Apparently, the solidity docs claim that even if an assembly block does not return to solidity it is not memory safe... So we might need to leave these annotations off which baffles me...

> Since this is mainly about the optimizer, these restrictions still need to be followed, even if the assembly block reverts or terminates. As an example, the following assembly snippet is not memory safe, because the value of returndatasize() may exceed the 64 byte scratch space:

https://docs.soliditylang.org/en/v0.8.16/assembly.html

src/curve/stable/StablePair.sol
```diff
@@ -127,7 +130,10 @@ contract StablePair is ReservoirPair {
     function burn(address) external override returns (uint256, uint256) {
         // DELEGATE TO StableMintBurn
         address lTarget = MINT_BURN_LOGIC;
-        assembly {
+
+        // SAFETY:
+        // This assembly block is memory safe as it does not return to solidity
+        assembly ("memory-safe") {
```
OliverNChalk: Ditto here

OliverNChalk (changes requested):
Comment about reducing the verboseness/length of safety messages + the fact that we might not be allowed to mark the blocks that return to solidity as safe even though they feel safe.

I would suggest we remove the non-returning memory-safe claims, then check the gas difference if we add them in on a branch

test/unit/ConstantProductPair.t.sol
xenide: I generated both assemblies and tried to diff them [here](https://github.com/vexchange/v3-core/tree/bug/via-ir/asm). Not sure if this way of diffing is the most useful, and what's the next step from here. 

In the diff file in that branch I've deleted a lot of things that are obviously not useful

OliverNChalk: lmao, that diff resulted in 8000 lines different? if you want i can look into this for you

src/GenericFactory.sol
OliverNChalk: can resolve

src/GenericFactory.sol
xenide: okay fixed

src/curve/stable/StablePair.sol
xenide: Okay so here's my life of logic. Correct me if my line of thought is wrong. 

What you quoted (solidity docs) serves as an example generally. HOWEVER, in our case in `StableMintBurn`, both calldata and returndata fit within 64 bytes as per my explanation [here](https://github.com/vexchange/v3-core/pull/150#discussion_r1081023287). 

So the reason that it is memory safe is that they both don't exceed 64 bytes, not that it doesn't release back to solidity. UNLESS, `returndatasize` does not respect `(uint256, uint256)` as specified in `StableMintBurn::burn`. We don't expect this to happen do we? 

^ is my line of logic above sensible?

OliverNChalk: That's true if lTarget == MINT_BURN_LOGIC which seems fairly likely we can control. I would write a comment to that effect

xenide: okay, have updated the comment

src/curve/stable/StablePair.sol
xenide: updated the comment as above

test/unit/ConstantProductPair.t.sol
xenide: Given that this bug doesn't affect our core code's correctness, it's not super urgent to me right now.

Honestly I want to contribute back to the solidity community and report this bug, but I don't know how to make sense of the diff of the assembly output for now. 

So I leave it up to you whether you want to investigate further and get the real diff of the opcodes. I don't mind filling the bug report once we have distilled the essence of the diff

OliverNChalk: Agreed

src/GenericFactory.sol
```diff
@@ -63,20 +66,22 @@ contract GenericFactory is IGenericFactory, Owned {
             address lPointer = lByteCode[i];
             uint256 lSize = lPointer.code.length - 0x01;
 
-            // TODO: Go check/annotate all asm for memory safety.
-            assembly {
+            // SAFETY:
+            // This assembly block is memory safe because the free memory pointer is updated after the write
```
OliverNChalk:
```suggestion
            // Updates the free memory pointer before returning.
```

src/GenericFactory.sol
```diff
@@ -63,20 +66,22 @@ contract GenericFactory is IGenericFactory, Owned {
...
             lFreeMem += lSize;
             // TODO: Do we need to pad to 32 bytes?
             lByteCodeLength += lSize;
         }
 
-        // TODO: Releasing back to solidity after the loop is not memory safe.
-        // Write the copied size & update free_mem.
-        assembly {
+        // SAFETY:
+        // This assembly block is memory safe as the free memory pointer is updated after the write
```
OliverNChalk:
```suggestion
        // Updates the free memory pointer after the write.
```

src/GenericFactory.sol
```diff
@@ -147,7 +152,9 @@ contract GenericFactory is IGenericFactory, Owned {
         // TODO: Test that _loadCurve errors for invalid indexes.
         bytes memory lInitCode = _loadCurve(aCurveId, lToken0, lToken1);
 
-        assembly {
+        // SAFETY:
+        // This assembly block is memory safe as it does not write to memory
```
OliverNChalk:
```suggestion
        // Does not write to memory
```

src/GenericFactory.sol
```diff
@@ -193,7 +200,10 @@ contract GenericFactory is IGenericFactory, Owned {
         require(_deployInProgress, "FACTORY: ONLY_CHILDREN_CAN_CALL");
 
         bytes memory lInitCode = getBytecode(aCodeKey, aToken0, aToken1);
-        assembly {
+
+        // SAFETY:
+        // This assembly block is memory safe as it does not write to memory
```
OliverNChalk:
```suggestion
        // Does not write to memory
```

src/curve/stable/StablePair.sol
```diff
@@ -127,7 +132,12 @@ contract StablePair is ReservoirPair {
     function burn(address) external override returns (uint256, uint256) {
         // DELEGATE TO StableMintBurn
         address lTarget = MINT_BURN_LOGIC;
-        assembly {
+
+        // SAFETY:
+        // This block is memory safe as as the delegated call has the same signature as the calling function
+        // and both the calldata and returndata do not exceed 64 bytes
+        // This is only valid when lTarget == MINT_BURN_LOGIC
```
OliverNChalk:
```suggestion
        // The delegated call has the same signature as the calling function
        // and both the calldata and returndata do not exceed 64 bytes
        // This is only valid when lTarget == MINT_BURN_LOGIC
```

OliverNChalk (approved):

src/GenericFactory.sol
xenide: fixed

src/GenericFactory.sol
xenide: fixed

src/GenericFactory.sol
xenide: fixed

src/curve/stable/StablePair.sol
xenide: fixed this and the block for `mint` as well

PR #151, opened 2023-01-05
fix: remove try catch and other fixes

### LINKED ISSUE #143: fix: remove try/catch from `AssetManagedPair`
`AssetManagerPair` should not make assumptions about the 3rd party protocol. It is the asset manager that should know about and handle error conditions arising from the 3rd party protocols. Since the `AssetManagedPair` is not replaceable but the asset manager is.  

For the first version of the AaveManager at least, we allow `afterLiquidityEvent` to fail if Aave goes into frozen / paused mode

xenide:
## Solution

- fixed some todos, the biggest being the change to make token0/1 ERC20. 
- would like to merge this first before the PR gets too big and unwieldy


src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -107,7 +107,6 @@ contract ConstantProductPair is ReservoirPair {
                 if (lSqrtNewK > lSqrtOldK) {
                     uint256 lSharesToIssue = _calcFee(lSqrtNewK, lSqrtOldK, platformFee, totalSupply);
 
-                    // TODO: What happens if no PLATFORM_FEE_TO is set
```
xenide, line 110: for this, some platformFee would be minted to the zero address. 

We can just make sure to set a `platformFeeTo` in the factory right from the start so that this wouldn't happen. Then we wouldn't need to bloat the curves' code with a check that it is non-zero

src/curve/stable/StableMintBurn.sol
```diff
@@ -109,7 +108,6 @@ contract StableMintBurn is ReservoirPair {
         // places)
         // which results in 112 + 60 + 1 = 173 bits
         // which fits into uint192
-        // TODO: Why does lastInvariant get set to lNewLiq? Not symmetric with burn?
```
xenide, line 112: we can set it to `lNewLiq` here cuz it is calculated when we read the balances. However for burn, we need to calculate it again after the tokens have been redeemed, as the balances have changed then.

src/curve/constant-product/ConstantProductPair.sol
OliverNChalk: Sounds good

src/curve/stable/StableMintBurn.sol
OliverNChalk: Got it

src/UniswapV2ERC20.sol
```diff
@@ -1,100 +0,0 @@
-/* solhint-disable const-name-snakecase */
-pragma solidity ^0.8.0;
-
-import { IUniswapV2ERC20 } from "src/interfaces/IUniswapV2ERC20.sol";
-
-// TODO: Is UniswapV2ERC20 still the best ERC-20 to inherit from or should we
```
xenide, line 6: indeed, solmate's one is more gas efficient (comparing to the existing one and openzeppelin's one) and interestingly is based on uni v2's impl. 

I've replaced it with such

src/ReservoirPair.sol
```diff
@@ -23,19 +25,15 @@ abstract contract ReservoirPair is AssetManagedPair, OracleWriter {
...
+    function _checkedTransfer(ERC20 aToken, address aDestination, uint256 aAmount, uint256 aReserve0, uint256 aReserve1)
+        internal
+    {
+        if (!_safeTransfer(address(aToken), aDestination, aAmount)) {
             uint256 tokenOutManaged = aToken == token0 ? token0Managed : token1Managed;
             uint256 reserveOut = aToken == token0 ? aReserve0 : aReserve1;
             if (reserveOut - tokenOutManaged < aAmount) {
                 assetManager.returnAsset(aToken == token0, aAmount - (reserveOut - tokenOutManaged));
-                require(_safeTransfer(aToken, aDestination, aAmount), "RP: TRANSFER_FAILED");
+                require(_safeTransfer(address(aToken), aDestination, aAmount), "RP: TRANSFER_FAILED");
```
OliverNChalk, line 36: Not make _safeTransfer take an ERC20?

src/curve/stable/StableMintBurn.sol
```diff
@@ -119,9 +117,8 @@ contract StableMintBurn is ReservoirPair {
         _managerCallback();
     }
 
-    // TODO: public -> external?
     /// @dev Burns LP tokens sent to this contract. The router must ensure that the user gets sufficient output tokens.
-    function burn(address aTo) public returns (uint256 amount0, uint256 amount1) {
+    function burn(address aTo) external returns (uint256 amount0, uint256 amount1) {
```
OliverNChalk, line 121: Do these changes need to be replicated to the main StablePair contract?

OliverNChalk (approved):
minor comments, looks good.

We said AaveManager would revert if Aave does so from this PoV behaviour looks good

src/ReservoirPair.sol
xenide: I considered this, but `_safeTransfer` needs to take in an `address` type in order to use the low level `.call()`. There's no way to do like `token0.transfer()` and make it return `(bool, bytes)` as far as I know

src/curve/stable/StableMintBurn.sol
xenide: yes good catch, 

fixed on the main `StablePair` contract

PR #153, opened 2023-01-07
refactor: collapse classes

### LINKED ISSUE #152: chore: combine `Pair`, `OracleWriter`, `AssetManagedPair`, `ReservoirPair` into one class
Name it probably `ReservoirPair`

xenide:
## Solution

- do we find a value in still having an `IPair` interface? 
   - an argument for it would be, we can put the natspec docs there instead of in the implementation code so that it looks neater and separated


OliverNChalk:
implementing interfaces is useless unless its an ERC interface where strict compliance is necessary. If there is no one using IPair, i'd say delete it

OliverNChalk (approved):

PR #154, opened 2023-01-09
fix: todos and some perf 

xenide:


src/curve/stable/StableMintBurn.sol
```diff
@@ -75,9 +75,8 @@ contract StableMintBurn is ReservoirPair {
 
     function mint(address aTo) external override returns (uint256 rLiquidity) {
         // NB: Must sync management PNL before we load reserves.
-        // TODO: Is passing/using reserves as uint256 cheaper and still safe?
```
xenide, line 78: from the diff from the changes made, it does seem a bit cheaper. Will follow up with a comment regarding safety

```diff
1c1
< AssetManagedPairTest:testAdjustManagement(uint256,uint256) (runs: 256, μ: 311670, ~: 311363)
---
> AssetManagedPairTest:testAdjustManagement(uint256,uint256) (runs: 256, μ: 311676, ~: 311644)
3,7c3,7
< AssetManagedPairTest:testAdjustManagement_DecreaseManagement(uint256,uint256) (runs: 256, μ: 498902, ~: 501910)
< AssetManagedPairTest:testAdjustManagement_KStillHolds(uint256) (runs: 256, μ: 694125, ~: 686608)
< AssetManagedPairTest:testBurn_AfterAlmostTotalLoss() (gas: 656146)
< AssetManagedPairTest:testBurn_AfterLoss(uint256,uint256) (runs: 256, μ: 700914, ~: 703526)
< AssetManagedPairTest:testMint_AfterLoss(uint256,uint256) (runs: 256, μ: 627106, ~: 627974)
---
> AssetManagedPairTest:testAdjustManagement_DecreaseManagement(uint256,uint256) (runs: 256, μ: 499068, ~: 501910)
> AssetManagedPairTest:testAdjustManagement_KStillHolds(uint256) (runs: 256, μ: 694043, ~: 686528)
> AssetManagedPairTest:testBurn_AfterAlmostTotalLoss() (gas: 656131)
> AssetManagedPairTest:testBurn_AfterLoss(uint256,uint256) (runs: 256, μ: 700912, ~: 703752)
> AssetManagedPairTest:testMint_AfterLoss(uint256,uint256) (runs: 256, μ: 627079, ~: 627934)
9,10c9,10
< AssetManagedPairTest:testSetManager_CannotMigrateWithManaged(uint256,uint256) (runs: 256, μ: 407575, ~: 407268)
< AssetManagedPairTest:testSwap_AfterLoss(uint256) (runs: 256, μ: 716804, ~: 712336)
---
> AssetManagedPairTest:testSetManager_CannotMigrateWithManaged(uint256,uint256) (runs: 256, μ: 407581, ~: 407549)
> AssetManagedPairTest:testSwap_AfterLoss(uint256) (runs: 256, μ: 716726, ~: 712258)
12,15c12,15
< AssetManagedPairTest:testSyncManaged_ConstantProduct(uint256,uint256) (runs: 256, μ: 396579, ~: 396512)
< AssetManagedPairTest:testSyncManaged_Stable(uint256,uint256) (runs: 256, μ: 367668, ~: 369890)
< ConstantProductPairTest:testBurn() (gas: 165259)
< ConstantProductPairTest:testBurn_Zero() (gas: 103331)
---
> AssetManagedPairTest:testSyncManaged_ConstantProduct(uint256,uint256) (runs: 256, μ: 396582, ~: 396515)
> AssetManagedPairTest:testSyncManaged_Stable(uint256,uint256) (runs: 256, μ: 367610, ~: 369832)
> ConstantProductPairTest:testBurn() (gas: 165262)
> ConstantProductPairTest:testBurn_Zero() (gas: 103334)
18,26c18,26
< ConstantProductPairTest:testMint_JustAboveMinimumLiquidity() (gas: 4813385)
< ConstantProductPairTest:testMint_MinimumLiquidity() (gas: 4766777)
< ConstantProductPairTest:testMint_UnderMinimumLiquidity() (gas: 4709122)
< ConstantProductPairTest:testOracle_ClampedPrice_AtLimit() (gas: 218954)
< ConstantProductPairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 218901)
< ConstantProductPairTest:testOracle_ClampedPrice_OverLimit() (gas: 219434)
< ConstantProductPairTest:testOracle_CorrectLiquidity() (gas: 304838)
< ConstantProductPairTest:testOracle_CorrectPrice() (gas: 313209)
< ConstantProductPairTest:testOracle_CorrectPriceDiffDecimals() (gas: 4909038)
---
> ConstantProductPairTest:testMint_JustAboveMinimumLiquidity() (gas: 4772240)
> ConstantProductPairTest:testMint_MinimumLiquidity() (gas: 4725632)
> ConstantProductPairTest:testMint_UnderMinimumLiquidity() (gas: 4667977)
> ConstantProductPairTest:testOracle_ClampedPrice_AtLimit() (gas: 218936)
> ConstantProductPairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 218883)
> ConstantProductPairTest:testOracle_ClampedPrice_OverLimit() (gas: 219416)
> ConstantProductPairTest:testOracle_CorrectLiquidity() (gas: 304841)
> ConstantProductPairTest:testOracle_CorrectPrice() (gas: 313173)
> ConstantProductPairTest:testOracle_CorrectPriceDiffDecimals() (gas: 4867893)
28c28
< ConstantProductPairTest:testOracle_NoWriteInSameTimestamp() (gas: 223949)
---
> ConstantProductPairTest:testOracle_NoWriteInSameTimestamp() (gas: 223934)
30,45c30,45
< ConstantProductPairTest:testOracle_OverflowAccPrice() (gas: 180556)
< ConstantProductPairTest:testOracle_SimplePrices() (gas: 309841)
< ConstantProductPairTest:testOracle_WrapsAroundAfterFull() (gas: 4310634689)
< ConstantProductPairTest:testPlatformFee_Disable() (gas: 423603)
< ConstantProductPairTest:testPlatformFee_DisableReenable() (gas: 638648)
< ConstantProductPairTest:testSwap() (gas: 83244)
< ConstantProductPairTest:testSwap_ExactOut(uint256) (runs: 256, μ: 93291, ~: 93022)
< ConstantProductPairTest:testSwap_ExactOutExceedReserves() (gas: 38155)
< ConstantProductPairTest:testSwap_ExactOut_NewReservesExceedUint104() (gas: 82318)
< ConstantProductPairTest:testSwap_ExtremeAmounts() (gas: 4879570)
< ConstantProductPairTest:testSwap_MinInt256() (gas: 14954)
< ConstantProductPairTest:testSwap_Reenter() (gas: 80334)
< ConstantProductPairTest:testWriteObservations() (gas: 245812)
< FlashSwapTest:testSwap_FlashSwap_ExactIn(uint256) (runs: 256, μ: 221297, ~: 225815)
< FlashSwapTest:testSwap_FlashSwap_ExactOut(uint256) (runs: 256, μ: 226434, ~: 226154)
< FlashSwapTest:testSwap_FlashSwap_NoPay(uint256) (runs: 256, μ: 180784, ~: 185368)
---
> ConstantProductPairTest:testOracle_OverflowAccPrice() (gas: 180538)
> ConstantProductPairTest:testOracle_SimplePrices() (gas: 309805)
> ConstantProductPairTest:testOracle_WrapsAroundAfterFull() (gas: 4309454969)
> ConstantProductPairTest:testPlatformFee_Disable() (gas: 423550)
> ConstantProductPairTest:testPlatformFee_DisableReenable() (gas: 638542)
> ConstantProductPairTest:testSwap() (gas: 83226)
> ConstantProductPairTest:testSwap_ExactOut(uint256) (runs: 256, μ: 93267, ~: 92998)
> ConstantProductPairTest:testSwap_ExactOutExceedReserves() (gas: 38179)
> ConstantProductPairTest:testSwap_ExactOut_NewReservesExceedUint104() (gas: 82294)
> ConstantProductPairTest:testSwap_ExtremeAmounts() (gas: 4838407)
> ConstantProductPairTest:testSwap_MinInt256() (gas: 14966)
> ConstantProductPairTest:testSwap_Reenter() (gas: 80322)
> ConstantProductPairTest:testWriteObservations() (gas: 245776)
> FlashSwapTest:testSwap_FlashSwap_ExactIn(uint256) (runs: 256, μ: 221231, ~: 225776)
> FlashSwapTest:testSwap_FlashSwap_ExactOut(uint256) (runs: 256, μ: 226382, ~: 226103)
> FlashSwapTest:testSwap_FlashSwap_NoPay(uint256) (runs: 256, μ: 180745, ~: 185329)
48,49c48,49
< GenericFactoryTest:testAllPairs() (gas: 14234938)
< GenericFactoryTest:testCreatePair_AllCurves(uint256) (runs: 256, μ: 7131455, ~: 7131278)
---
> GenericFactoryTest:testAllPairs() (gas: 14142789)
> GenericFactoryTest:testCreatePair_AllCurves(uint256) (runs: 256, μ: 6987945, ~: 4591165)
51,54c51,54
< GenericFactoryTest:testCreatePair_IdenticalAddress(uint256) (runs: 256, μ: 19960, ~: 19961)
< GenericFactoryTest:testCreatePair_MoreThan18Decimals(uint256) (runs: 256, μ: 269918, ~: 269740)
< GenericFactoryTest:testCreatePair_PairAlreadyExists(uint256) (runs: 256, μ: 24479, ~: 24480)
< GenericFactoryTest:testCreatePair_ZeroAddress(uint256) (runs: 256, μ: 19983, ~: 19984)
---
> GenericFactoryTest:testCreatePair_IdenticalAddress(uint256) (runs: 256, μ: 19965, ~: 19961)
> GenericFactoryTest:testCreatePair_MoreThan18Decimals(uint256) (runs: 256, μ: 269188, ~: 258338)
> GenericFactoryTest:testCreatePair_PairAlreadyExists(uint256) (runs: 256, μ: 24484, ~: 24480)
> GenericFactoryTest:testCreatePair_ZeroAddress(uint256) (runs: 256, μ: 19988, ~: 19984)
62,64c62,64
< OracleWriterTest:testOracle_SamePriceDiffLiq() (gas: 14767495)
< OracleWriterTest:testOracle_SamePriceSameLiq() (gas: 14766018)
< OracleWriterTest:testOracle_SameReservesDiffPrice() (gas: 14767931)
---
> OracleWriterTest:testOracle_SamePriceDiffLiq() (gas: 14675318)
> OracleWriterTest:testOracle_SamePriceSameLiq() (gas: 14673841)
> OracleWriterTest:testOracle_SameReservesDiffPrice() (gas: 14675754)
73c73
< PairTest:testEmitEventOnCreation() (gas: 14233108)
---
> PairTest:testEmitEventOnCreation() (gas: 14140959)
87,90c87,90
< ReservoirPairTest:testCheckedTransfer_RevertWhenTransferFail() (gas: 139149)
< ReservoirPairTest:testCheckedTransfer_RevertWhenTransferReverts() (gas: 395772)
< ReservoirPairTest:testOracleWriteAfterAssetManagerProfit_Burn() (gas: 785201)
< ReservoirPairTest:testOracleWriteAfterAssetManagerProfit_Mint() (gas: 833888)
---
> ReservoirPairTest:testCheckedTransfer_RevertWhenTransferFail() (gas: 139122)
> ReservoirPairTest:testCheckedTransfer_RevertWhenTransferReverts() (gas: 395751)
> ReservoirPairTest:testOracleWriteAfterAssetManagerProfit_Burn() (gas: 785186)
> ReservoirPairTest:testOracleWriteAfterAssetManagerProfit_Mint() (gas: 833848)
94,103c94,103
< StablePairTest:testAttackWhileRampingDown_LongInterval() (gas: 223537)
< StablePairTest:testAttackWhileRampingDown_ShortInterval() (gas: 222941)
< StablePairTest:testBurn() (gas: 137371)
< StablePairTest:testBurn_DiffDecimalPlaces(uint256) (runs: 256, μ: 9891011, ~: 9890571)
< StablePairTest:testBurn_LastInvariantUseReserveInsteadOfBalance() (gas: 229530)
< StablePairTest:testBurn_Reenter() (gas: 51945)
< StablePairTest:testBurn_WhenRampingA(uint256) (runs: 256, μ: 421799, ~: 421925)
< StablePairTest:testBurn_Zero() (gas: 74651)
< StablePairTest:testFactoryAmpTooHigh() (gas: 659286)
< StablePairTest:testFactoryAmpTooLow() (gas: 654287)
---
> StablePairTest:testAttackWhileRampingDown_LongInterval() (gas: 223498)
> StablePairTest:testAttackWhileRampingDown_ShortInterval() (gas: 222902)
> StablePairTest:testBurn() (gas: 137353)
> StablePairTest:testBurn_DiffDecimalPlaces(uint256) (runs: 256, μ: 9839972, ~: 9839521)
> StablePairTest:testBurn_LastInvariantUseReserveInsteadOfBalance() (gas: 229473)
> StablePairTest:testBurn_Reenter() (gas: 51957)
> StablePairTest:testBurn_WhenRampingA(uint256) (runs: 256, μ: 421763, ~: 421889)
> StablePairTest:testBurn_Zero() (gas: 74633)
> StablePairTest:testFactoryAmpTooHigh() (gas: 659164)
> StablePairTest:testFactoryAmpTooLow() (gas: 654165)
105,119c105,119
< StablePairTest:testMint() (gas: 110443)
< StablePairTest:testMintFee_DiffPlatformFees(uint256) (runs: 256, μ: 11154909, ~: 11158301)
< StablePairTest:testMintFee_WhenRampingA_PoolBalanced(uint256) (runs: 256, μ: 11341648, ~: 11341777)
< StablePairTest:testMintFee_WhenRampingA_PoolUnbalanced(uint256) (runs: 256, μ: 10781829, ~: 10781348)
< StablePairTest:testMint_NonOptimalProportion() (gas: 143108)
< StablePairTest:testMint_NonOptimalProportion_ThenBurn() (gas: 296403)
< StablePairTest:testMint_OnlyTransferOneToken() (gas: 9669819)
< StablePairTest:testMint_PlatformFeeOff() (gas: 115946)
< StablePairTest:testMint_Reenter() (gas: 51902)
< StablePairTest:testMint_WhenRampingA(uint256) (runs: 256, μ: 444652, ~: 444781)
< StablePairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 240138)
< StablePairTest:testOracle_CorrectLiquidity() (gas: 283980)
< StablePairTest:testOracle_CorrectPrice() (gas: 366165)
< StablePairTest:testOracle_LiquidityAtMaximum() (gas: 273902)
< StablePairTest:testOracle_NoWriteInSameTimestamp() (gas: 212343)
---
> StablePairTest:testMint() (gas: 110403)
> StablePairTest:testMintFee_DiffPlatformFees(uint256) (runs: 256, μ: 11103079, ~: 11106471)
> StablePairTest:testMintFee_WhenRampingA_PoolBalanced(uint256) (runs: 256, μ: 11289800, ~: 11289929)
> StablePairTest:testMintFee_WhenRampingA_PoolUnbalanced(uint256) (runs: 256, μ: 10730329, ~: 10729860)
> StablePairTest:testMint_NonOptimalProportion() (gas: 143068)
> StablePairTest:testMint_NonOptimalProportion_ThenBurn() (gas: 296327)
> StablePairTest:testMint_OnlyTransferOneToken() (gas: 9618851)
> StablePairTest:testMint_PlatformFeeOff() (gas: 115906)
> StablePairTest:testMint_Reenter() (gas: 51914)
> StablePairTest:testMint_WhenRampingA(uint256) (runs: 256, μ: 444572, ~: 444701)
> StablePairTest:testOracle_ClampedPrice_NoDiffWithinLimit() (gas: 240120)
> StablePairTest:testOracle_CorrectLiquidity() (gas: 283962)
> StablePairTest:testOracle_CorrectPrice() (gas: 366123)
> StablePairTest:testOracle_LiquidityAtMaximum() (gas: 273862)
> StablePairTest:testOracle_NoWriteInSameTimestamp() (gas: 212304)
121,125c121,125
< StablePairTest:testOracle_OverflowAccPrice() (gas: 197420)
< StablePairTest:testOracle_SimplePrices() (gas: 402012)
< StablePairTest:testOracle_WrapsAroundAfterFull() (gas: 5430733140)
< StablePairTest:testPlatformFee_Disable() (gas: 445814)
< StablePairTest:testPlatformFee_DisableReenable() (gas: 645032)
---
> StablePairTest:testOracle_OverflowAccPrice() (gas: 197402)
> StablePairTest:testOracle_SimplePrices() (gas: 401970)
> StablePairTest:testOracle_WrapsAroundAfterFull() (gas: 5429356800)
> StablePairTest:testPlatformFee_Disable() (gas: 445723)
> StablePairTest:testPlatformFee_DisableReenable() (gas: 644850)
134,135c134,135
< StablePairTest:testRampA_SwappingDuringRampingDown(uint256,uint256,uint256,uint256) (runs: 256, μ: 561310, ~: 576913)
< StablePairTest:testRampA_SwappingDuringRampingUp(uint256,uint256,uint256,uint256) (runs: 256, μ: 561344, ~: 577049)
---
> StablePairTest:testRampA_SwappingDuringRampingDown(uint256,uint256,uint256,uint256) (runs: 256, μ: 561226, ~: 576829)
> StablePairTest:testRampA_SwappingDuringRampingUp(uint256,uint256,uint256,uint256) (runs: 256, μ: 561255, ~: 576965)
140,155c140,155
< StablePairTest:testSwap() (gas: 87442)
< StablePairTest:testSwap_BetterPerformanceThanConstantProduct() (gas: 128789)
< StablePairTest:testSwap_DiffAs(uint256,uint256,uint256) (runs: 256, μ: 9910246, ~: 9911425)
< StablePairTest:testSwap_DiffSwapFees(uint256) (runs: 256, μ: 9914640, ~: 9917051)
< StablePairTest:testSwap_ExactInExceedUint104() (gas: 79343)
< StablePairTest:testSwap_ExactOutExceedReserves() (gas: 38085)
< StablePairTest:testSwap_IncreasingSwapFees(uint256,uint256,uint256) (runs: 256, μ: 305625, ~: 305624)
< StablePairTest:testSwap_MinInt256() (gas: 14986)
< StablePairTest:testSwap_Reenter() (gas: 90393)
< StablePairTest:testSwap_Token0ExactOut(uint256) (runs: 256, μ: 123098, ~: 131146)
< StablePairTest:testSwap_Token1ExactOut(uint256) (runs: 256, μ: 123233, ~: 131281)
< StablePairTest:testSwap_VeryLargeLiquidity(uint256) (runs: 256, μ: 9883039, ~: 9882472)
< StablePairTest:testSwap_VerySmallLiquidity(uint256,uint256,uint256) (runs: 256, μ: 9880968, ~: 9880790)
< StablePairTest:testSwap_ZeroInput() (gas: 14502)
< StablePairTest:testWriteObservations() (gas: 280279)
< ConstantProductPairGas:testGasBurn() (gas: 123163)
---
> StablePairTest:testSwap() (gas: 87421)
> StablePairTest:testSwap_BetterPerformanceThanConstantProduct() (gas: 128750)
> StablePairTest:testSwap_DiffAs(uint256,uint256,uint256) (runs: 256, μ: 9859187, ~: 9860375)
> StablePairTest:testSwap_DiffSwapFees(uint256) (runs: 256, μ: 9863590, ~: 9866001)
> StablePairTest:testSwap_ExactInExceedUint104() (gas: 79322)
> StablePairTest:testSwap_ExactOutExceedReserves() (gas: 38109)
> StablePairTest:testSwap_IncreasingSwapFees(uint256,uint256,uint256) (runs: 256, μ: 305562, ~: 305561)
> StablePairTest:testSwap_MinInt256() (gas: 14998)
> StablePairTest:testSwap_Reenter() (gas: 90378)
> StablePairTest:testSwap_Token0ExactOut(uint256) (runs: 256, μ: 123074, ~: 131119)
> StablePairTest:testSwap_Token1ExactOut(uint256) (runs: 256, μ: 123209, ~: 131254)
> StablePairTest:testSwap_VeryLargeLiquidity(uint256) (runs: 256, μ: 9831986, ~: 9831419)
> StablePairTest:testSwap_VerySmallLiquidity(uint256,uint256,uint256) (runs: 256, μ: 9829903, ~: 9829740)
> StablePairTest:testSwap_ZeroInput() (gas: 14514)
> StablePairTest:testWriteObservations() (gas: 280237)
> ConstantProductPairGas:testGasBurn() (gas: 123166)
157,160c157,160
< ConstantProductPairGas:testGasMint_Initial() (gas: 204575)
< ConstantProductPairGas:testGasSwap() (gas: 57629)
< ConstantProductPairGas:testGasSwap_UpdateOracle() (gas: 115108)
< ConstantProductPairGas:testGasSwap_UpdateOracleClamped() (gas: 110374)
---
> ConstantProductPairGas:testGasMint_Initial() (gas: 204587)
> ConstantProductPairGas:testGasSwap() (gas: 57611)
> ConstantProductPairGas:testGasSwap_UpdateOracle() (gas: 115090)
> ConstantProductPairGas:testGasSwap_UpdateOracleClamped() (gas: 110356)
167,173c167,173
< StablePairGas:testGasBurn() (gas: 97364)
< StablePairGas:testGasMint() (gas: 118901)
< StablePairGas:testGasMint_Initial() (gas: 221688)
< StablePairGas:testGasSwap() (gas: 67179)
< StablePairGas:testGasSwap_UpdateOracle() (gas: 132184)
< StablePairGas:testGasSwap_UpdateOracleClamped() (gas: 128149)
< ConstantProductMathTest:testCalcLogPrice_ReturnsOneWeiWhenPriceDiffGreaterThan1e18(uint256,uint256) (runs: 256, μ: 8432, ~: 8210)
---
> StablePairGas:testGasBurn() (gas: 97346)
> StablePairGas:testGasMint() (gas: 118861)
> StablePairGas:testGasMint_Initial() (gas: 221648)
> StablePairGas:testGasSwap() (gas: 67158)
> StablePairGas:testGasSwap_UpdateOracle() (gas: 132163)
> StablePairGas:testGasSwap_UpdateOracleClamped() (gas: 128128)
> ConstantProductMathTest:testCalcLogPrice_ReturnsOneWeiWhenPriceDiffGreaterThan1e18(uint256,uint256) (runs: 256, μ: 8434, ~: 8210)
175,178c175,178
< ConstantsLibTest:testMintBurnKey() (gas: 11190)
< StableMathTest:testComputeLiquidityFromAdjustedBalances_ConvergeEvenWithVeryUnbalancedValues(uint256,uint256,uint256) (runs: 256, μ: 18707, ~: 13675)
< StableMathTest:testGetAmountIn(uint256,uint256,uint256) (runs: 256, μ: 24566, ~: 24075)
< StableMathTest:testGetAmountOut(uint256,uint256,uint256) (runs: 256, μ: 27288, ~: 26461)
---
> ConstantsLibTest:testMintBurnKey() (gas: 11160)
> StableMathTest:testComputeLiquidityFromAdjustedBalances_ConvergeEvenWithVeryUnbalancedValues(uint256,uint256,uint256) (runs: 256, μ: 18640, ~: 13613)
> StableMathTest:testGetAmountIn(uint256,uint256,uint256) (runs: 256, μ: 24573, ~: 24075)
> StableMathTest:testGetAmountOut(uint256,uint256,uint256) (runs: 256, μ: 27286, ~: 26461)
``

File: src/ReservoirPair.sol [hunk 2]
```diff
@@ -336,7 +339,6 @@ abstract contract ReservoirPair is ReservoirERC20 {
         returns (uint104 rReserve0, uint104 rReserve1)
     {
         if (address(assetManager) == address(0)) {
-            // PERF: Is assigning to rReserve0 cheaper?
```
xenide, line 339: attempted assigning to return value and introducing an else for the rest of the procedure. 

There was no difference at all for gas so this is prolly optimized by the compiler + optimizer already

src/curve/stable/StablePair.sol
```diff
@@ -275,7 +275,6 @@ contract StablePair is ReservoirPair {
     }
 
     /// @dev number of coins in the pool multiplied by A precise
-    // perf: is it possible to optimize/simplify by hardcoding to two assets instead of using _getNA() etc
```
xenide, line 276: This is done already since we already hardcoded it to 2

src/asset-management/AaveManager.sol
```diff
@@ -12,8 +12,6 @@ import { IPoolAddressesProvider } from "src/interfaces/aave/IPoolAddressesProvid
 import { IPool } from "src/interfaces/aave/IPool.sol";
 import { IAaveProtocolDataProvider } from "src/interfaces/aave/IAaveProtocolDataProvider.sol";
 
-// TODO: Should probably move the adapter away from AssetManagedPair because
```
xenide, line 15: may I know what did you mean by this? Is it about not using an `AssetManagedPair` interface?

src/curve/stable/StableMintBurn.sol
xenide: one difference we might want to address is that `_mintFee` for `StablePair` takes in a wider type (uint256) while `ConstantProductPair` takes an `uint104`. This taking in of a narrower type is inherited from UniV2. Not sure why they did it that way, the args don't get packed do they? 

Safety wise: 
There could be a scenario where, when we first load the reserves, it still fits within uint104, but when we `_syncManaged`, the additional profits pushes it above `uint104`. But this will be taken care of by solidity's builtin checked arithmetic. So the whole tx will revert if this happens. 

Casting the reserves for `_updateAndUnlock` should not be a problem since it should still fit within `uint104`, if it hasn't overflowed from `_syncManaged`. So this is safe. 

My recommendation is: 
1. to change `ConstantProductPair::_mintFee` to take in `uint256` instead of `uint104` since it is going to be cast into `uint256` later anyway.

OliverNChalk: Yes the recommendation makes sense, we should align both and revert if size exceeds reserves (as we do)

src/asset-management/AaveManager.sol
OliverNChalk: idk anymore, can resolve

src/ReservoirPair.sol
```diff
@@ -262,18 +261,21 @@ abstract contract ReservoirPair is ReservoirERC20 {
...
-    /// @param data calls to with this data, in the event of a flash swap
-    function swap(int256 amount, bool inOrOut, address to, bytes calldata data) external virtual returns (uint256 amountOut);
+    /// @param aAmount positive to indicate token0, negative to indicate token1
+    /// @param aInOrOut true to indicate exact amount in, false to indicate exact amount out
+    /// @param aTo address to send the output token and leftover input tokens, callee for the flash swap
+    /// @param aData calls to with this data, in the event of a flash swap
+    function swap(int256 aAmount, bool aInOrOut, address aTo, bytes calldata aData)
+        external
+        virtual
+        returns (uint256 amountOut);
```
OliverNChalk: Should be `rAmountOut`

src/curve/stable/StableMintBurn.sol
OliverNChalk: If we are going to make `_lockAndLoad` return a uint256 (for gas savings) then maybe we should make `updateAndUnlock` accept a uint256 and perform the cast, reverting on overflow?

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -118,18 +120,18 @@ contract ConstantProductPair is ReservoirPair {
     }
 
     function mint(address aTo) external override returns (uint256 rLiquidity) {
-        (uint104 lReserve0, uint104 lReserve1, uint32 lBlockTimestampLast,) = _lockAndLoad();
-        (lReserve0, lReserve1) = _syncManaged(lReserve0, lReserve1); // check asset-manager pnl
+        (uint256 lReserve0, uint256 lReserve1, uint32 lBlockTimestampLast,) = _lockAndLoad();
+        (lReserve0, lReserve1) = _syncManaged(uint104(lReserve0), uint104(lReserve1)); // check asset-manager pnl
```
OliverNChalk, line 124: The point of using uint256 is to avoid the cost of casting, we lose this saving if we immediately cast. _syncManaged should accept uint256, the only place we should perform (safe) casting is in `_lockAndLoad()` and `updateAndUnlock()`

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -118,18 +120,18 @@ contract ConstantProductPair is ReservoirPair {
...
 
         uint256 lBalance0 = _totalToken0();
         uint256 lBalance1 = _totalToken1();
         uint256 lAmount0 = lBalance0 - lReserve0;
         uint256 lAmount1 = lBalance1 - lReserve1;
 
-        _mintFee(lReserve0, lReserve1);
+        _mintFee(uint104(lReserve0), uint104(lReserve1));
```
OliverNChalk: Similarly

test/unit/ConstantProductPair.t.sol
```diff
@@ -432,7 +433,7 @@ contract ConstantProductPairTest is BaseTest, IReservoirCallee {
             LogCompression.fromLowResLog(
                 (lObs2.logAccRawPrice - lObs0.logAccRawPrice) / int32(lObs2.timestamp - lObs0.timestamp)
             ),
-            Math.sqrt(lPrice1 * lPrice2),
```
OliverNChalk: I'd move the `Math` library into `test/` and use it for the assertions. Prove no regression when we change libraries (for instance if we find an even more performant sqrt in solmate or solady etc)

OliverNChalk: Please add issue to benchmark with `solmate` and `solady` sqrt

test/unit/StablePair.t.sol
```diff
@@ -1499,7 +1524,7 @@ contract StablePairTest is BaseTest {
             LogCompression.fromLowResLog(
                 (lObs2.logAccRawPrice - lObs0.logAccRawPrice) / int32(lObs2.timestamp - lObs0.timestamp)
             ),
-            Math.sqrt(lPrice1 * lPrice2),
+            FixedPointMathLib.sqrt(lPrice1 * lPrice2),
```
OliverNChalk: I would ban usage of FixedPointMathLib in test/ (at least for math functions we can replicate other ways, like sqrt)

OliverNChalk (changes requested):
Most looks good but the `uint104 -> uint256` feels like a step back unless we do it properly (use uint256 everywhere except where we load/store it

test/unit/ConstantProductPair.t.sol
xenide: okay created #155

src/ReservoirPair.sol

src/ReservoirPair.sol
```diff
@@ -331,23 +336,22 @@ abstract contract ReservoirPair is ReservoirERC20 {
...
+        uint256 lToken0Managed = assetManager.getBalance(this, token0);
+        uint256 lToken1Managed = assetManager.getBalance(this, token1);
 
         rReserve0 = _handleReport(token0, aReserve0, token0Managed, lToken0Managed);
         rReserve1 = _handleReport(token1, aReserve1, token1Managed, lToken1Managed);
 
-        token0Managed = lToken0Managed;
-        token1Managed = lToken1Managed;
+        token0Managed = lToken0Managed.toUint104();
+        token1Managed = lToken1Managed.toUint104();
```
xenide, line 356: I think it's appropriate to do a safecast over here since there's no guarantee that the new balance will be <= uint104

src/ReservoirPair.sol
```diff
@@ -357,32 +361,32 @@ abstract contract ReservoirPair is ReservoirERC20 {
...
-        if (token0Change > 0) {
-            uint104 lDelta = uint104(uint256(int256(token0Change)));
+        if (aToken0Change > 0) {
+            uint104 lDelta = uint104(uint256(aToken0Change));
             token0Managed += lDelta;
             token0.transfer(msg.sender, lDelta);
-        } else if (token0Change < 0) {
-            uint104 lDelta = uint104(uint256(int256(-token0Change)));
+        } else if (aToken0Change < 0) {
+            uint104 lDelta = uint104(uint256(-aToken0Change));
```
xenide: Albeit there is no guarantee that the `aToken0/1Change` would be <= uint104, I think it would be reasonable to assume that they would be, since it is delta, and not the entire amount. 

What do you think? Do we need a safecast here as well?

src/ReservoirPair.sol
```diff
@@ -357,32 +361,32 @@ abstract contract ReservoirPair is ReservoirERC20 {
         assetManager.afterLiquidityEvent();
     }
 
-    function adjustManagement(int256 token0Change, int256 token1Change) external {
+    function adjustManagement(int256 aToken0Change, int256 aToken1Change) external {
         require(msg.sender == address(assetManager), "AMP: AUTH_NOT_MANAGER");
-        require(token0Change != type(int256).min && token1Change != type(int256).min, "AMP: CAST_WOULD_OVERFLOW");
+        require(aToken0Change != type(int256).min && aToken1Change != type(int256).min, "AMP: CAST_WOULD_OVERFLOW");
```
xenide: technically we don't need this check, since the cast would overflow if `type(int256).min` is supplied. 

I know cuz I tried it

OliverNChalk: If you cast int256.min to uint256 do you get a big number? or how/where does this get caught?

src/ReservoirPair.sol
OliverNChalk: Delta is almost as big as entire amount, 50% of uint128 is still > uint127, (remember each bit of data doubles/halves the number)

src/ReservoirPair.sol
```diff
@@ -357,32 +361,32 @@ abstract contract ReservoirPair is ReservoirERC20 {
...
             token0.transfer(msg.sender, lDelta);
-        } else if (token0Change < 0) {
-            uint104 lDelta = uint104(uint256(int256(-token0Change)));
+        } else if (aToken0Change < 0) {
+            uint104 lDelta = uint104(uint256(-aToken0Change));
 
             // solhint-disable-next-line reentrancy
             token0Managed -= lDelta;
 
             token0.transferFrom(msg.sender, address(this), lDelta);
```
OliverNChalk: Should we be using safeTransferFrom everywhere?

src/ReservoirPair.sol
xenide: it gets caught at 

```
uint104 lDelta = uint104(uint256(-aToken0Change));
```

Specifically at the negating operation. When we negate `aToken0Change`, it attempts to assign a value in the same type, but with opposing signs. And since `type(int256).max` (....9967) is 1 less than `type(int256).min` (-...9968), this casting will fail. 

For the positive case, the cast will always succeed without problems.

src/ReservoirPair.sol
xenide: good thinking... I'll prolly use solmate's impl

src/ReservoirPair.sol
xenide: mmm yeah I'm keenly aware of that. So I guess a safecast is warranted here?

src/ReservoirPair.sol
xenide: implemented

src/asset-management/AaveManager.sol
```diff
@@ -53,18 +49,17 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev returns the balance of the token managed by various markets in the native precision
-    function getBalance(ReservoirPair aOwner, ERC20 aToken) external view returns (uint104 rTokenBalance) {
+    function getBalance(ReservoirPair aOwner, ERC20 aToken) external view returns (uint256 rTokenBalance) {
         return _getBalance(aOwner, aToken);
     }
 
-    function _getBalance(ReservoirPair aOwner, ERC20 aToken) private view returns (uint104 rTokenBalance) {
+    function _getBalance(ReservoirPair aOwner, ERC20 aToken) private view returns (uint256 rTokenBalance) {
```
xenide, line 58: I have reviewed AaveManager as a whole and concluded that changing the types to `uint256` will not cause any detrimental effects within the manager itself. 

The manager is self-contained, we just have to be careful when the pair queries balances from the manager

OliverNChalk: Okay but how do users get their money back if `_getBalance` returns `type(uint256).max`, is there a way to lower the balances?

xenide: realistically to hit `type(uint256).max`, the entire supply of the ERC20 token would have to be here, managed by our asset manager. So that wouldn't happen right, you agree? 

Putting this problem in the more likely scenario of the balance exceeding `uint104` (the entire reserves being managed by AM), the AM admin can call `adjustManagement` to get it under, cuz it would make `token0/1Managed` overflow `uint104` as well. But, that still doesn't solve the problem cuz after redeeming, it would make the reserves exceed `uint104` as well. I'll write a test case to demonstrate this problem first. 

It seems like we need a solution similar to `skim` (for `token0/1Managed` instead of reserves), to act as a release valve when the amount gets crazy.

test/unit/StablePair.t.sol
xenide: okay I replaced the test usages of sqrt to use the previously used Babylonian sqrt implementation

src/asset-management/AaveManager.sol
OliverNChalk: Yeah, right now this is a major/critical vuln

xenide: wrote a test case to demonstrate this scenario happening. 

Still contemplating what's the best solution for it

src/ReservoirPair.sol
OliverNChalk: Are you sure the cast will fail, that isnt what i would expect. I thought solidity casts are unchecked?

src/ReservoirPair.sol
OliverNChalk: Yeah

src/ReservoirPair.sol
xenide: Technically it isn't casting that fails, it's the negation and assigning of a value that doesn't fit. If you only cast it without negating, it will not fail. I'm sure of this because I've written tests in my own repo to test it out. 

You can run it yourself [here](https://github.com/xenide/balancer-log-math/blob/881860c9260c320e9f1fbf64a48808b50516edf6/test/Counter.t.sol#L43). This behavior only started in 0.8.0 onwards.

src/ReservoirPair.sol
xenide: okay implemented with tests

src/ReservoirPair.sol
OliverNChalk: Interesting, ill update my understanding haha. So safeCast is only needed for uint to uint or int to int casts? (or rather casts of different sizes)?

xenide: correct `SafeCast` is most useful when you're casting from a wider type to a narrower type.  

When casting from an `int` to a `uint`, and vice versa, we better check the sign. Although there're functions such as `toUint256(int256 value)` and `toInt256(uint256 value)` in the lib, the onus is still on us to check the sign since it reverts if the types don't fit. 

In our case, since we check if `token0Change >/< 0 `, our unchecked casting to `uint256` is **safe**. 

Therefore, I've deleted the `require` statement.

src/asset-management/AaveManager.sol
xenide: will address this with a solution in the subsequent PR

OliverNChalk: Can resolve if this is captured in a separate issue or the PR is started

src/ReservoirPair.sol
```diff
@@ -38,6 +40,8 @@ struct Observation {
 abstract contract ReservoirPair is ReservoirERC20 {
     using FactoryStoreLib for GenericFactory;
     using Bytes32Lib for bytes32;
+    using SafeCast for uint256;
+    using SafeTransferLib for ERC20;
 
     event SwapFeeChanged(uint256 oldSwapFee, uint256 newSwapFee);
     event CustomSwapFeeChanged(uint256 oldCustomSwapFee, uint256 newCustomSwapFee);
```
OliverNChalk, line 47:
```
    string internal constant PLATFORM_FEE_TO_NAME = "Shared::platformFeeTo";
    string private constant PLATFORM_FEE_NAME = "Shared::platformFee";
    string private constant RECOVERER_NAME = "Shared::defaultRecoverer";
    bytes4 private constant SELECTOR = bytes4(keccak256("transfer(address,uint256)"));
```

SELECTOR is a poor name, if this is used can we rename it to `ERC20_TRANSFER`

For the `*_NAME` we can drop the `_NAME` I think because its implied by the type

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -87,23 +89,23 @@ contract ConstantProductPair is ReservoirPair {
         // ASSERT: newK & oldK        < uint104
         // ASSERT: aPlatformFee       < FEE_ACCURACY
         // ASSERT: aCirculatingShares < uint104
+        unchecked {
```
OliverNChalk, line 92: We added unchecked here because it was in the original formula?

OliverNChalk (commented):
Code changes look good but its hard for me to rationalize the gas snapshot diff:

image

OliverNChalk (approved):

src/ReservoirPair.sol
xenide: fixed

src/curve/constant-product/ConstantProductPair.sol
xenide: we added an unchecked here cuz your comments (perf) said it can be unchecked.

So I made an assessment and concluded that it is indeed safe to make it unchecked

xenide:
> Code changes look good but its hard for me to rationalize the gas snapshot diff:
> 

I think this gas savings is due to using the more optimized `sqrt` implementation by solmate, which uses assembly 

Crazy I know...

src/asset-management/AaveManager.sol
xenide: will fix in #156

PR #158, opened 2023-01-21
benchmark: solmate vs solady

### LINKED ISSUE #155: test: benchmark `solmate` and `solady` impls

xenide:
## Solution

- for `FixedPointMathLib`: 
   - after isolating this change, it seems that solady is a net positive for gas efficiency. Red for solmate, green for solady 

```diff
< ConstantProductPairTest:testBurn() (gas: 129947)
< ConstantProductPairTest:testMint() (gas: 97956)
< ConstantProductPairGas:testGasBurn() (gas: 87161)
< ConstantProductPairGas:testGasMint() (gas: 94667)
< ConstantProductPairGas:testGasMint_Initial() (gas: 186853)
< StablePairGas:testGasBurn() (gas: 95121)
< StablePairGas:testGasMint() (gas: 116373)
< StablePairGas:testGasMint_Initial() (gas: 211330)
< StablePairGas:testGasSwap() (gas: 64616)
< StablePairGas:testGasSwap_UpdateOracle() (gas: 109043)
< StablePairGas:testGasSwap_UpdateOracleClamped() (gas: 105010)
---
> ConstantProductPairTest:testBurn() (gas: 129739)
> ConstantProductPairTest:testMint() (gas: 97748)
> ConstantProductPairGas:testGasBurn() (gas: 86953)
> ConstantProductPairGas:testGasMint() (gas: 94459)
> ConstantProductPairGas:testGasMint_Initial() (gas: 186735)
> StablePairGas:testGasBurn() (gas: 95095)
> StablePairGas:testGasMint() (gas: 116347)
> StablePairGas:testGasMint_Initial() (gas: 211217)
> StablePairGas:testGasSwap() (gas: 64603)
> StablePairGas:testGasSwap_UpdateOracle() (gas: 108846)
> StablePairGas:testGasSwap_UpdateOracleClamped() (gas: 104813)

```

- for `SSTORE2`:
   - for factory, solady's `SSTORE2` certainly seems to be more gas efficient

```diff
< GenericFactoryTest:testAddCurve() (gas: 116634)
< GenericFactoryTest:testAllPairs() (gas: 13216698)
< GenericFactoryTest:testCreatePair_AllCurves(uint256) (runs: 256, μ: 6603641, ~: 4286660)
---
> GenericFactoryTest:testAddCurve() (gas: 116040)
> GenericFactoryTest:testAllPairs() (gas: 13215898)
> GenericFactoryTest:testCreatePair_AllCurves(uint256) (runs: 256, μ: 6603244, ~: 4286660)
```

- for `safeTransfer` + `safeApprove`, solady seems like a win too: 

```diff
-StablePairGas:testGasBurn() (gas: 95121)
-StablePairGas:testGasMint() (gas: 116373)
-StablePairGas:testGasMint_Initial() (gas: 211330)
+StablePairGas:testGasBurn() (gas: 95095)
+StablePairGas:testGasMint() (gas: 116347)
+StablePairGas:testGasMint_Initial() (gas: 211217)
```

- don't think solady's `Ownable` is suitable as the functions they expose are quite a number more than solmate's, and quite unnecessary


OliverNChalk:
Just a heads up, `forge snapshot --diff` is a million times more readable than a diff like this. If you have already updated the `.gas-snapshot`file what you can do is:

```bash
git checkout <commit-before-optimizations> .gas-snapshot # Reset to old snapshot
forge snapshot --diff # Read/save the comparison
git checkout .gas-snapshot # Restore the snapshot
```

src/ReservoirPair.sol
```diff
@@ -41,7 +41,7 @@ abstract contract ReservoirPair is ReservoirERC20 {
     using FactoryStoreLib for GenericFactory;
     using Bytes32Lib for bytes32;
     using SafeCast for uint256;
-    using SafeTransferLib for ERC20;
+//    using SafeTransferLib for ERC20;
```
OliverNChalk: Remove the comments before merging

src/asset-management/AaveManager.sol
```diff
@@ -117,13 +117,15 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
         uint256 lShares = _decreaseShares(aPair, aToken, aAaveToken, aAmount);
         pool.withdraw(address(aToken), aAmount, address(this));
         emit FundsDivested(aPair, aToken, lShares);
-        aToken.safeApprove(address(aPair), aAmount);
+//        aToken.safeApprove(address(aPair), aAmount);
+        SafeTransferLib.safeApprove(address(aToken), address(aPair), aAmount);
```
OliverNChalk, line 119: Are `solmate` and `solady` safeTransfer/safeApprove equivalent? E.g. perform all the same checks?

src/libraries/StableOracleMath.sol
```diff
@@ -1,7 +1,7 @@
 // TODO: License, this was taken from Balancer right?
 pragma solidity ^0.8.0;
 
-import { FixedPointMathLib } from "solmate/utils/FixedPointMathLib.sol";
+import { FixedPointMathLib } from "solady/utils/FixedPointMathLib.sol";
```
OliverNChalk, line 4: Would be good to benchmark our `sqrt` implementation (can also be in another PR).

src/asset-management/AaveManager.sol
xenide: `solmate`'s `safeTransfer`, `safeTransferFrom`, and `safeApprove`, checks if the `returndatasize` is greater than 31 bytes, `solady`'s doesn't. I guess this is to guard against implementations that return something but not exactly a `bool` or full 32 bytes? But they check that the return value is EXACTLY `1` so I think solmate's additional check is not very necessary?

Both impl checks if the `returndatasize` is zero.

I'd imagine both implementations would have problems with `BNB`'s [implementation](https://etherscan.io/address/0xb8c77482e45f1f44de1745f52c74426c631bdd52#code) of `transfer` since it doesn't return anything?

src/libraries/StableOracleMath.sol
xenide: I kinda tested it previously when addressing the comment to use a more efficient `Math` lib. Our naive lib with the Babylon impl is definitely a lot less efficient than theirs

src/ReservoirPair.sol

src/libraries/StableOracleMath.sol
OliverNChalk: are we using solady sqrt or balancer sqrt?

src/asset-management/AaveManager.sol
OliverNChalk: if it does not return anything then it is assumed to succeed, if it does reutrn something then you need to check if its a bool

OliverNChalk (approved):

src/asset-management/AaveManager.sol
xenide: so are you saying that you agree with `solmate`'s additional check for a `returndatasize` greater than 31 bytes?  

Cuz the return value could be 1, and the size is greater than 0, but the type could be `int`, `uint`, `uint8` etc...

src/libraries/StableOracleMath.sol
xenide: eh? long time ago we were using balancer's Babylonian method. Then we switched to solmate's which was a lot more efficient. Now it seems that solady's one is even more efficient. 

But if you're talking about comparing the correctness, we have the our unit tests using the Babylonian impl, checking against the src's use of whatever impl of `sqrt`

OliverNChalk: Sounds good, happy to resolve

src/asset-management/AaveManager.sol
OliverNChalk: Hmm, I'd want to know what use case solamte's extra check is covering. Though, I'm happy to use solady for now and have a task which states "answer why solmate checks data size". Which can be a low prio back log task

xenide: tracked in #159, resolving for now

PR #160, opened 2023-01-26
fix: reintroduce `IAssetManager` interface

xenide:
## Solution

- indeed solves the compilation problem for repos that import `v3-core` as a submodule


xenide:
disabling `via-ir` for now as it is causing some issues: 

1. non-deterministic output. Running it locally vs on github CI somehow gives different compilation results, which can result in the failure of the tests (hash of `StableMintBurn`)
2. And even if the tests work on this repo (as a submodule), when the periphery repo's tests import `BaseTest`, the constructor running will still fail, even if periphery has `via-ir` enabled. 
   - I've tried to debug this for a bit but have no idea what is causing this 
      - I've made sure that the 3rd party lib versions of `v3-core` and `v3-periphery` are exactly the same

PR #162, opened 2023-02-16
fix: StablePair creation gas too high

xenide:


xenide:
the diff at this point in time, after avoiding one call to factory in `StableMintBurn`: 

```diff
testCreateStablePair() (gas: -1289 (-0.014%))
```

PR #163, opened 2023-02-18
fix: `StableMintBurn` deploy once

xenide:
## Motivation

This is to address the issue of gas usage exceeding the block gas limit in AVAX (8M gas) when deploying the StablePair. It was due to having to deploy 2 contracts. Now we re-use `StableMintBurn` for all StablePairs, leading to only one contract of it being used for `delegatecall` by all StablePairs. So a deployment of a `StablePair` is only about 4.8M gas

## Solution

- I was thinking of ways we can further optimize this:  
  - make StableMintBurn inherit from a storage layout base class that has the same layout as `StablePair`, instead of from `ReservoirPair`
     - we can still go ahead with this so that the constructor is not needed? Cuz right now we're still calling the constructor but it has no material meaning 
  - there's no way to avoid making `token0` and `token1` mutable is there? Cuz in the end the StablePair `delegatecall` has to read StablePair's storage anyway, so they HAVE to be in storage instead of code


src/libraries/Constants.sol
```diff
@@ -4,5 +4,5 @@ pragma solidity ^0.8.0;
 import { StableMintBurn } from "src/curve/stable/StableMintBurn.sol";
 
 library ConstantsLib {
-    bytes32 public constant MINT_BURN_KEY = bytes32(0xfca7c68f711102d77ab0a2765cee66cea90b7f07334352820e7ebf31c0ef3acf);
+    bytes32 public constant MINT_BURN_KEY = bytes32(0xf629f8de34ecb99dffcf69ea812ed1fdbba96bee14757da972c41f5b986fc74d);
```
OliverNChalk: This should become a constant address instead of a key

src/curve/stable/StablePair.sol
```diff
@@ -52,8 +52,8 @@ contract StablePair is ReservoirPair {
     uint64 private lastInvariantAmp;
 
     constructor(address aToken0, address aToken1) ReservoirPair(aToken0, aToken1, PAIR_SWAP_FEE_NAME) {
-        MINT_BURN_LOGIC = factory.deploy(ConstantsLib.MINT_BURN_KEY, aToken0, aToken1);
-        require(MINT_BURN_LOGIC != address(0), "SP: MINT_BURN_DEPLOYMENT_FAILED");
+        MINT_BURN_LOGIC = factory.read("SP::STABLE_MINT_BURN").toAddress();
+        require(address(MINT_BURN_LOGIC).code.length > 0, "SP: MINT_BURN_NOT_DEPLOYED");
```
OliverNChalk: This should become a constant address instead of a key. We only used the key because we needed to deploy the contract's bytecode from a key. However, now that we have a single contract the address should be constant across all networks/deployments/stable-pairs

src/GenericFactory.sol
```diff
@@ -195,22 +191,16 @@ contract GenericFactory is IGenericFactory, Owned {
...
-            // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3ac4add548178708f5401c26280b952beb244c1e/contracts/utils/Create2.sol#L40
+        assembly("memory-safe") {
             rContract := create2(callvalue(), add(lInitCode, 0x20), mload(lInitCode), 0)
 
-            if iszero(extcodesize(rContract)) { revert(0, 0) }
+            if iszero(extcodesize(rContract)) { revert(0,0) }
         }
-
         emit Deployed(aCodeKey, rContract);
     }
```
OliverNChalk: What happened here with the formatting?

- The space between assembly and memory-safe
- The newline before emitting the event
- The spacing on revert

OliverNChalk (commented):
Some key changes that need to happen:

- Retain immutable vars, gas is too juicy
- Replace stable-mint-burn-key with stable-mint-burn-address
- Explore inheriting stable pair (and just override all public functions to be empty except the ones we actually use, so we can still deploy StableMintBurn).

src/GenericFactory.sol
xenide: my bad, I was doing this by hand and actually had another function. 

Anyway, just fixed it.

src/GenericFactory.sol
```diff
@@ -7,13 +7,28 @@ import { SSTORE2 } from "solady/utils/SSTORE2.sol";
...
+        // Does not write to memory
+        assembly ("memory-safe") {
+            // sanity checked against OZ implementation:
+            // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3ac4add548178708f5401c26280b952beb244c1e/contracts/utils/Create2.sol#L40
+            lStableMintBurn := create2(callvalue(), add(lInitCode, 0x20), mload(lInitCode), 0)
+
+            if iszero(extcodesize(lStableMintBurn)) { revert(0, 0) }
+        }
+        emit Deployed(lStableMintBurn);
+    }
```
xenide: actually, we can just use the solidity syntax for this instead of the assembly one right ,since we don't have to do the memory manipulation anymore?

OliverNChalk: Correct. We can just do `new StableMintBurn{salt: bytes32(0)}();`

src/curve/stable/StableMintBurn.sol
```diff
@@ -3,55 +3,68 @@
...
-    uint64 initialATime;
-    /// @dev futureATime is a unix timestamp and will only overflow every 584 billion years
-    uint64 futureATime;
-}
+import { ReservoirERC20, ERC20 } from "src/ReservoirERC20.sol";
+import { Slot0, Observation } from "src/ReservoirPair.sol";
+import { StablePair, AmplificationData } from "src/curve/stable/StablePair.sol";
 
-contract StableMintBurn is ReservoirPair {
+contract StableMintBurn is ReservoirERC20, IAssetManagedPair {
```
OliverNChalk, line 22: Do we need to implement the interface?

src/curve/stable/StableMintBurn.sol
```diff
@@ -246,4 +382,40 @@ contract StableMintBurn is ReservoirPair {
...
+                rClampedPrice = aPrevClampedPrice * (1e18 - (maxChangeRate * aTimeElapsed)) / 1e18;
+            }
+            rClampedLogPrice = int112(LogCompression.toLowResLog(rClampedPrice));
+        } else {
+            rClampedPrice = aCurrRawPrice;
+            rClampedLogPrice = int112(LogCompression.toLowResLog(aCurrRawPrice));
+        }
+    }
+
+    function adjustManagement(int256 aToken0Change, int256 aToken1Change) external { }
```
OliverNChalk, line 414: Can we just drop these empty functions?

src/curve/stable/StablePair.sol
```diff
@@ -44,17 +42,13 @@ contract StablePair is ReservoirPair {
 
     AmplificationData public ampData;
 
-    uint256 private _locked = 1;
-
     // We need the 2 variables below to calculate the growth in liquidity between
```
OliverNChalk, line 45: Good spot, weird we didn't notice - guess we don't have solhint running?

OliverNChalk (approved):
Minor comments but this looks good to me. We can do a final comb over before audit trying to reduce duplication but happy with this as now

src/GenericFactory.sol
xenide: okay done

src/curve/stable/StablePair.sol
xenide: so I just tried, and solhint doesn't catch this. It's like the unused imports, solhint doesn't catch it currently. 

Also I just learned that there's a solhint rule that recommends not more than having 15 state variables

OliverNChalk: Interesting, ty for checking

src/curve/stable/StableMintBurn.sol
xenide: we need to implement the interface cuz `IAssetManager` requires a `IAssetManagedPair` in the `getBalance` function. I considered the alternative, which is to change the interface, then we'll have to change a lot of existing code, which is really annoying and we lose the strong type

src/curve/stable/StableMintBurn.sol
xenide: because we need to inherit the interface, we need to have these blank functions, else it wouldn't compile

PR #164, opened 2023-02-19
feat: make `MINT_BURN_LOGIC` fixed address and `StableMintBurn` standalone address 

xenide:
## Solution

- so if we make `StableMintBurn` inherit from `StablePair`, then it is not possible to hard code the `MINT_BURN_LOGIC` as a `constant` within `StablePair` itself, cuz it would be a circular loop. Right now it's using the dynamic calculation but that's too expensive in prod. So it seems that a read from the factory is still needed
- there is no way to avoid calling the constructor / providing valid token0/1 addresses to `StableMintBurn` is there? If that's the case, then we have to pass in a valid ERC20 token for it to call, else the constructor fails. 
   - one way to overcome this is to `CREATE2` two dummy tokens on each network, and then supply those as constructor arguments
   - perhaps the "cleanest" way to do it, is just to make `StableMintBurn` not inherit from anyone, but make it have the same storage layout as `StablePair`. That way, there is no need to call the constructor but at the same time allows us to `delegatecall`
- I've indeed verified using `forge inspect --pretty [StablePair | StableMintBurn] storage` that both contracts' storage layout is exactly the same 
- if only one copy of `StableMintBurn` is needed, then we don't need to add its bytecode, so I've removed it so we can just deploy straight up

### DIFF
--- a/src/GenericFactory.sol
+++ b/src/GenericFactory.sol
@@ -185,19 +185,17 @@ contract GenericFactory is IGenericFactory, Owned {
 
     event Deployed(address _address);
 
-    function deploySharedContract(bytes calldata aInitCode, address aToken0, address aToken1)
+    function deploySharedContract(bytes memory aInitCode)
         external
         onlyOwner
         returns (address rContract)
     {
-        bytes memory lInitCode = bytes.concat(aInitCode, abi.encode(aToken0), abi.encode(aToken1));
-
         // SAFETY:
         // Does not write to memory
         assembly ("memory-safe") {
             // sanity checked against OZ implementation:
             // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3ac4add548178708f5401c26280b952beb244c1e/contracts/utils/Create2.sol#L40
-            rContract := create2(callvalue(), add(lInitCode, 0x20), mload(lInitCode), 0)
+            rContract := create2(callvalue(), add(aInitCode, 0x20), mload(aInitCode), 0)
 
             if iszero(extcodesize(rContract)) { revert(0, 0) }
         }
--- a/src/GenericFactory.sol
+++ b/src/GenericFactory.sol
@@ -185,11 +185,7 @@ contract GenericFactory is IGenericFactory, Owned {
 
     event Deployed(address _address);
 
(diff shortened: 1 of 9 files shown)


OliverNChalk:
Your logic seems correct to me, happy to strip the contract down and just manually ensure the storage layouts match

xenide:
so I managed to do it. But now it seems that we might have to make a decision:

Current approach (`StableMintBurn` as a standalone contract)
- Pros: 
  - `StableMintBurn` address can be a constant and that address baked into all `StablePairs` for all chains
- Cons: 
  - a lot of duplicate code (400 including blank lines) which may introduce maintainability + security issues when we change code but forget to update it for all instances 
  - assuming 300 lines of code that would be ~$9K of additional cost for audit 

Make StablePair read `StableMintBurn` address from `factory`, and `StableMintBurn` inherit from `StablePair`
- Pros:
   - almost no duplicated code, no maintainability issues 
   - no additional cost for audit 
- Cons: 
   - gotta do a call to factory to read the address (which costs a few hundred gas at most)
   - address might be diff for each chain, so we need to keep track / maintain that
      - one way to overcome this is to modify the constructor of `ReservoirPair`, such that if both tokens are `address-zero`, we treat it differently, and then we just pass two `address-zero` args into the `StableMintBurn` constructor

xenide:
so a few hours into this work, I realized that the approach we agreed on (making `StableMintBurn` inherit from `StablePair` has its issues), namely: 
- all the internal functions inherited from `StablePair` and `ReservoirPair` still use the immutable `token0/1`. Which means, we gotta import + override ALL of them, which makes it the same as the current approach (standalone). 
- unless we make `token0/1` mutable again, then we won't have this problem 

So now it's a choice between spending more $ and having duplicated code vs making it more expensive to do all operations for all curve types. I guess we'll still choose the former?

OliverNChalk:
I suggest go duplicate code approach no inheritance. We should essentially just replicate the storage layout then move mint/burn and their dependencies into the stable pair. Hopefully genuine duplication should be kept to a minimum (we can also move math/helper funcs into libs if there are any shared ones).

For this PR just get it working with duplication then we can consider a refactor later for audit cost/maintainability purposes

PR #165, opened 2023-02-21
feat: SMT solver

xenide:
## Motivation

Progress so far: 
- on github CI, it complains that it needs the following, not sure how do have it installed for the CI
`warning[7649]: Warning: CHC analysis was not possible since no Horn solver was found and enabled. The accepted solvers for CHC are Eldarica and z3.
[27](https://github.com/vexchange/v3-core/actions/runs/4230990648/jobs/7348967238#step:4:28)` 

- when run locally, it takes around 2-3 hours and then ends with: 

```
Error:
Solc Error:
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
         with:
           version: nightly
       - run: sudo apt install libz3-4 -y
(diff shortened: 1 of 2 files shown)


OliverNChalk:
image

Should be able to just sudo apt install z3 -y

xenide:
hmm I tried installing z3 / libz3, and even tried to sym link the file it's looking for with the actual `.so` file. Still complains. Not sure how to proceed

PR #166, opened 2023-02-27
fix: make stable invariant public

xenide:
## Solution

- there is no need to change the visibility of these two variables for `StableMintBurn`. In fact, if we changed them to public, it would cost more to deploy them. Which makes me think, I can prolly change ALL of `StableMintBurn`'s variables to private


OliverNChalk (approved):
interesting re stablemintburn private, i agree

PR #168, opened 2023-03-07
fix: various todos

xenide:


src/GenericFactory.sol
```diff
@@ -146,7 +146,6 @@ contract GenericFactory is IGenericFactory, Owned {
 
         (address lToken0, address lToken1) = _sortAddresses(aTokenA, aTokenB);
 
-        // TODO: Test that _loadCurve errors for invalid indexes.
```
xenide, line 149: this is taken care of by `testCreatePair_CurveDoesNotExist`

src/GenericFactory.sol
```diff
@@ -79,7 +79,6 @@ contract GenericFactory is IGenericFactory, Owned {
             }
 
             lFreeMem += lSize;
-            // TODO: Do we need to pad to 32 bytes?
```
xenide, line 82: this should be fine as it has been working without problems and from our understanding of constructor bytecode

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -1,5 +1,4 @@
-// TODO: Can we reduce the nesting by deleting the parent dir?
```
xenide, line 1: I think the nesting makes it clearer which contract is for which curve. But don't have a super strong opinion of this tho

xenide:
@OliverNChalk can you address the two fallback `TODO`s?

src/GenericFactory.sol
OliverNChalk: agreed, can resolve

src/curve/constant-product/ConstantProductPair.sol
OliverNChalk: Yeah lets leave as is, this was before we split out StablePair from StableMintBurn

OliverNChalk:
If we convert to fallback we no longer implement ReservoirPair abstract mathoeds so instead de-duped the code. We should have a call at some point to discuss whether it is possible to have StableMintBurn extend ReservoirPair to minimize code dupe.

OliverNChalk (approved):

PR #170, opened 2023-03-09
refactor: make StableMintBurn inherit from ReservoirPair

OliverNChalk:


src/ReservoirPair.sol
```diff
@@ -100,8 +100,6 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
             : 0;
         swapFeeName = keccak256(abi.encodePacked(aSwapFeeName));
 
-        // TODO: What is the security risk if someone calls these methods on the
```
xenide: they can call and StableMintBurn's storage will be updated. But there is no risk as `swapFee`, `platformFee`, and `oracleCaller` are all read from `StablePair`'s storage

xenide:
in the latest commit I brought our idea to its logical conclusion. The advantage is that there is no more duplicated code. The only disadvantage is that `StablePair` has to read from the factory for the address of `StableMintBurn` instead of being embedded as a constant

PR #171, opened 2023-03-09
test: e2e deployment

xenide:


xenide:
Okay so for some reason, when we deploy a `MintableERC20` in anvil, it costs 8M+ gas which exceeds the gas limit, and then causes the script to fail if I configure anvil's gas limit to be 8M? 

When I run the script in the sdk repo forking AVAX mainnet, it's only 800K gas.

xenide:
right now the script runs, but when I introduce a gas limit of 8M in anvil, it gives an error of intrinsic gas too high. Not sure why. But the tx with the highest gas use is the factory, which uses about 6.25M gas

OliverNChalk:
> right now the script runs, but when I introduce a gas limit of 8M in anvil, it gives an error of intrinsic gas too high. Not sure why. But the tx with the highest gas use is the factory, which uses about 6.25M gas

The factory deploys the StableMintBurn contract from its constructor

xenide:
> > right now the script runs, but when I introduce a gas limit of 8M in anvil, it gives an error of intrinsic gas too high. Not sure why. But the tx with the highest gas use is the factory, which uses about 6.25M gas
> 
> The factory deploys the StableMintBurn contract from its constructor

I'm aware of that. That factory deployment tx costs `6254788` gas (which is inclusive of the `StableMintBurn` deployment), which IS lower than the block gas limit, which is why it baffles me

OliverNChalk:
gas too **high** is also weird. EVM only forwards 63/64 of the gas so maybe that's making the tx gas limit higher than the used gas.

OR forge is adding too much gas buffer. Are you able to force the tx gas limit to be 9_000_000?

xenide:
raised the gas limit to 8.5M and it works. Remaining questions: 

1. what other things do we need in the script? minting some liq? swapping? 
2. if we want to put this e2e testing as a ci step, can we use the default forge `msg.sender` / `tx.origin` instead of the custom one? Cuz it will affect the factory deploy address and thus the `StableMintBurn` address in `Constants.sol`. I'm trying to avoid another patching thing like what we have for the coverage

OliverNChalk:
Use this to run e2e tests `test/run_all_e2e.sh`:
```solidity
#!/bin/bash
set -euxo pipefail

for f in test/e2e/*.sh; do
  bash "$f"
done
```

Be sure to mark it executable `chmod +x test/run_all_e2e.sh`.

For `Q1` deployment should not to mint burn but these can be added if valuable. I'd do it in a later PR, I think this is fine for now.
For `Q2` im happy to use default if that's possible? I.e. do we have privkey for default sender?

xenide:
okay done with the additional script. 

For the second problem, since we're:
1. unable to use the private key of the default forge sender address
2. currently the factory is being deployed by the `msg.sender` manually do a create2 in `BaseTest`

Does it mean that we probably need to do that `CREATE2_DEPLOY_RAW` for BaseTest as well? And then get the test contract to create2 the factory with `msg.sender` as the owner. This will make sure that both e2e and unit tests have the same factory (bytecode) with the same owner, and therefore the same address

xenide:
managed to make it work by referring to `IGenericFactory` instead of `GenericFactory`. 

For the e2e testing script, since it's being broadcasted, we cannot do `vm.prank`, so the way to make it work is to set the guardians to the private key's corresponding wallet address (already done). But doing so would make unit tests fail, because the unit tests don't use that address. Unless we change the unit tests' `msg.sender` to use that address too

foundry.toml
```diff
@@ -9,7 +9,7 @@ match_path = "test/unit/*.sol"
 verbosity = 3
 fs_permissions = [
     { access = "write", path = "./scripts/optimized-deployer-meta" },
-    { access = "write", path = "./scripts/unoptimized-deployer-meta" },
+    { access = "write", path = "./scripts/unoptimized-deployer-meta" }
```
OliverNChalk: prefer trailing comma

OliverNChalk (approved):

foundry.toml
xenide: fixed

PR #172, opened 2023-03-14
fix: skim excess when amount managed exceeds `uint104.max`

xenide:


src/ReservoirPair.sol
```diff
@@ -385,6 +385,13 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
         }
     }
 
+    function _skimExcessManaged(ERC20 aToken, uint256 aAmtManaged) internal returns (uint256 rAmtSkimmed) {
+        address lRecoverer = factory.read(RECOVERER_NAME).toAddress();
+        rAmtSkimmed = aAmtManaged - type(uint104).max;
+        assetManager.returnAsset(aToken == _token0(), rAmtSkimmed);
+        _safeTransfer(address(aToken), lRecoverer, rAmtSkimmed);
```
xenide: I noticed, we should replace uni v2's impl of `_safeTransfer` with a lib implementation, yeah?

OliverNChalk: Yep, sounds good

OliverNChalk:
```suggestion
        address lRecoverer = factory.read(RECOVERER_NAME).toAddress();

        rAmtSkimmed = aAmtManaged - type(uint104).max;

        assetManager.returnAsset(aToken == _token0(), rAmtSkimmed);
        _safeTransfer(address(aToken), lRecoverer, rAmtSkimmed);
```

checks effects separation

src/ReservoirPair.sol
```diff
@@ -399,11 +406,25 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
+        if (lToken0Managed > type(uint104).max) {
+            uint256 lAmtSkimmed = _skimExcessManaged(lToken0, lToken0Managed);
+            lToken0Managed -= lAmtSkimmed;
+            aReserve0 -= lAmtSkimmed;
+        }
+        if (lToken1Managed > type(uint104).max) {
+            uint256 lAmtSkimmed = _skimExcessManaged(lToken1, lToken1Managed);
+            lToken1Managed -= lAmtSkimmed;
+            aReserve1 -= lAmtSkimmed;
+        }
```
OliverNChalk: I think it's better to revert if this fails then force this extra code/change right?

OliverNChalk: This is in line with how uni_v2 handles reserves overflowing

src/asset-management/AaveManager.sol
```diff
@@ -43,7 +43,13 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
         pool = IPool(addressesProvider.getPool());
-        dataProvider = IAaveProtocolDataProvider(addressesProvider.getPoolDataProvider());
+        updateDataProvider();
+    }
+
+    function updateDataProvider() public onlyOwner {
+        address lNewDataProvider = addressesProvider.getPoolDataProvider();
+        require(lNewDataProvider != address(0), "AM: DATA_PROVIDER_ADDRESS_ZERO");
+        dataProvider = IAaveProtocolDataProvider(lNewDataProvider);
     }
```
OliverNChalk: Interesting, did you find out that data provider can change over time?

xenide: I understood from aave's [docs](https://docs.aave.com/developers/core-contracts/pooladdressesprovider) that 

1. the `PoolAddressesProvider` will never change
2. that the admins can set the many of the addresses such as `ACLManager`, `PriceOracle` etc

It seems that I may have incorrectly assumed that the `Pool` address will also never change just by virtue that it is a proxy contract. They even mentioned that it's recommended to fetch the pool address each time you need to interact with it. But I think we don't have to do an external call every time. I'll add a function for the admin to update the Pool address as well.

OliverNChalk: Sounds good

xenide: alright, this I added

src/ReservoirPair.sol
xenide: okay for this I realize I cannot replace all instances of it. For this particular instance I can. 

But for `_checkedTransfer`, we need to get a return `bool` and not revert, which the library does not do. `try/catch` also doesn't work with the lib cuz it's assembly and not recognized as an external function call. So it seems that we have to keep our `_safeTransfer` for now

src/ReservoirPair.sol
xenide: you mean it's better to revert if the amount of managed tokens exceed `uint104`? Then the pair will brick and people can't even `burn`

src/ReservoirPair.sol

OliverNChalk: Try catch not working is not because its assembly, its because its an internal call. _safeTransfer does not use `CALL` it uses `JUMP`. Once it jumps to the internal function, that function then uses `CALL`, so you cannot catch it from a layer up sadly.

src/ReservoirPair.sol
OliverNChalk: How so, you can just skim can't you?

OliverNChalk: Skim excess should be publicly callable like normal skim

src/ReservoirPair.sol
OliverNChalk: Mixing different types of _safeTransfer scares me a little. Is it only inside of checkedTransfer that we use _safeTransfer? If so, can we just inline the call and delete _safeTransfer?

src/ReservoirPair.sol
xenide: okay I get what you mean now. It's good not to complicate the existing `_syncManaged` function and make it separate. 

I've done the implementation and tests as such

src/ReservoirPair.sol
xenide: yes it's only `_checkedTransfer` that uses `_safeTransfer` now. But there's **two** instances so if we inline we'll be repeating code?

OliverNChalk: Okay, lets not inline

OliverNChalk (approved):

PR #173, opened 2023-03-14
feat: reservoir deployer

OliverNChalk:


src/ReservoirPair.sol
```diff
@@ -55,8 +55,7 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
 
     string internal constant PLATFORM_FEE_TO_NAME = "Shared::platformFeeTo";
     string private constant PLATFORM_FEE_NAME = "Shared::platformFee";
-    // TODO: Rename from defaultRecoverer given its always read on recovery?
```
xenide: okay so this is renamed

src/ReservoirDeployer.sol
```diff
@@ -69,8 +69,7 @@ contract ReservoirDeployer {
         // Set global parameters.
         factory.write("Shared::platformFee", DEFAULT_PLATFORM_FEE);
         factory.write("Shared::platformFeeTo", address(this));
-        // TODO: Is it okay to not set defaultRecoverer?
```
xenide: let's just set it for now, cuz besides `recoverToken`, `skimExcessManaged` also relies on the recoverer. If we set it, it gives a more complete infra for testing

src/ReservoirDeployer.sol
```diff
@@ -104,7 +103,6 @@ contract ReservoirDeployer {
         step += 1;
     }
 
-    // TODO: Do we want to deploy an oracleCaller as part of standard deployment?
```
xenide: let's just deploy it so we have more comprehensive testing infra

src/ReservoirDeployer.sol
```diff
@@ -0,0 +1,189 @@
...
+        }
+        require(lFactoryAddress != address(0), "FAC_STEP: DEPLOYMENT_FAILED");
+
+        // Write the factory address so we can start configuring it.
+        factory = GenericFactory(lFactoryAddress);
+
+        // Set global parameters.
+        factory.write("Shared::platformFee", DEFAULT_PLATFORM_FEE);
+        factory.write("Shared::platformFeeTo", address(this));
+        factory.write("Shared::recoverer", msg.sender);
```
xenide: do we want to set these two OUTSIDE of the deployer? Cuz we may not know which addresses these are at the time of factory deployment. We certainly don't want the `platformFeeTo` to be the deployer do we?

src/ReservoirDeployer.sol
OliverNChalk: There is an inconsistency in skimExcessManaged. It does not check the recoverer is set while recoverToken does check this...

However, we really want skimExcessManaged to never fail as its a safety hatch. This makes me think we should force recoverer to always be non-zero and then remove the recoverToken check?

src/ReservoirDeployer.sol
OliverNChalk: If you mentally rename "deployer" to "crosschain multisig", does it then seem fine to set these?

xenide: yes if it's a cross-chain multi-sig then it's fine

src/ReservoirDeployer.sol
xenide: I realized that too, I'm okay with removing the check and we always make sure that there's a recoverer in the deploy scripts

src/ReservoirDeployer.sol
OliverNChalk: Okay, that's essentially what the guardians + rawCall enable

src/ReservoirDeployer.sol
OliverNChalk: I guess this is the best we can do short of make `recoverer` a custom config type that cannot be set to zero. I'm okay with this

xenide: okay check removed

foundry.toml
```diff
@@ -8,9 +8,10 @@ remappings = ["@openzeppelin/=lib/openzeppelin-contracts/contracts/"]
 match_path = "test/unit/*.sol"
 verbosity = 3
 fs_permissions = [
-    { access = "write", path = "./scripts/optimized-stable-mint-burn-address" },
-    { access = "write", path = "./scripts/unoptimized-stable-mint-burn-address" },
+    { access = "write", path = "./scripts/optimized-deployer-meta" },
+    { access = "write", path = "./scripts/unoptimized-deployer-meta" },
 ]
+ignored_error_codes=[]
```
xenide: saw this on [twitter](https://twitter.com/0x796/status/1638614525673873414) so I implemented it and caught several license issues

OliverNChalk: looks good, just spacing is wrong

scripts/unoptimized-deployer-meta
```diff
@@ -1,6 +1,6 @@
 {
-  "factory_hash": "0x8e3dc1fb25976a8279ce40a4b2c3d182158365da7c579cde433f5d8164691bf2",
-  "constant_product_hash": "0x6e3b13293eee4f98834bcde1f85c460a11a9323ce73efc080957044db4044c21",
-  "stable_hash": "0x3ca08204c301d01a2a143e86060aa7e8bcd991205dda2369d3b8899f9844b0c2",
+  "factory_hash": "0x542fe795ae57ecd8fb2233c98ae521ab2973ea1e286e679be44f573b1ec571a4",
+  "stable_hash": "0x4ac793675e33c3fe47b307e7ef3f14f35f5855c5f86244356b39f011be1de722",
+  "constant_product_hash": "0x0d0f6c6e5b513a7dfc8380dc45d750502c8a69b9122c62bcb967dff7a749b83f",
```
OliverNChalk: now deterministic: https://github.com/foundry-rs/foundry/issues/4592

Pushing the latest version

xenide (approved):

PR #176, opened 2023-03-25
feat: asset manager shutdown replacement 

### LINKED ISSUE #142: feat: AssetManager shutdown / replacement mode
redeem from 3rd party only, so that we can replace it with another / upgraded AssetManager

xenide:
## Motivation

- in this wind down mode we are introducing, we still want the `afterLiquidityEvent` call that attempts to increase the amount managed to succeed, cuz we don't want the mints/burns to fail 

## Solution

- questions: 
   - do we want some granularity of control for wind down? E.g. we set some pairs to wind down while doing normal operations for other pairs


OliverNChalk (approved):
Looks good

PR #177, opened 2023-03-25
feat: manage additional rewards in `AaveManager`

### LINKED ISSUE #161: fix(`AaveManager`): handle WAVAX rewards on top of base rewards
Currently the asset manager only tracks the shares of aave tokens each pair has. But we need a mechanism to handle the WAVAX rewards, and possibly to sell them and convert them into those aave tokens.

xenide:


src/asset-management/AaveManager.sol
```diff
@@ -193,6 +212,21 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
+    function claimRewardForMarket(address aMarket, address aReward) external returns (uint256 rClaimed) {
+        require(msg.sender == rewardSeller, "AM: NOT_REWARD_SELLER");
+        require(aReward != address(0), "AM: REWARD_TOKEN_ZERO");
+        require(aMarket != address(0), "AM: MARKET_ZERO");
+
+        address[] memory lMarkets = new address[](1);
+        lMarkets[0] = aMarket;
+
+        rClaimed = rewardsController.claimRewards(lMarkets, type(uint256).max, rewardSeller, aReward);
+    }
```
OliverNChalk, line 245: Once the reward seller performs the sale, how do the proceeds get distributed? We should ideally have a test case where the reward seller claims rewards, then swaps, then sends the result in. The swap can be faked by just minting the output asset.

xenide: okay I wrote a test case for this, also makes sure that everyone's shares increases proportionally

OliverNChalk (approved):

src/asset-management/AaveManager.sol
OliverNChalk: Nice looks good, approved

PR #180, opened 2023-04-04
fix: replace asserts with require if still necessary

### LINKED ISSUE #113: chore: remove all `assert`s and replace with `require` if still necessary

xenide:


src/libraries/ConstantProductOracleMath.sol
```diff
@@ -38,7 +38,6 @@ library ConstantProductOracleMath {
         spotPrice = reserve1.divWadUp(reserve0);
 
         int256 rawResult = LogCompression.toLowResLog(spotPrice);
-        assert(rawResult >= type(int112).min && rawResult <= type(int112).max);
```
xenide, line 41: removed these as the output from `LogCompression.toLowResLog` is guaranteed to be within range

src/libraries/StableOracleMath.sol
```diff
@@ -18,7 +18,6 @@ library StableOracleMath {
         spotPrice = calcSpotPrice(amplificationParameter, reserve0, reserve1);
 
         int256 rawLogSpotPrice = LogCompression.toLowResLog(spotPrice);
-        assert(rawLogSpotPrice >= type(int112).min && rawLogSpotPrice <= type(int112).max);
```
xenide, line 21: as above

src/libraries/ConstantProductOracleMath.sol
```diff
@@ -48,7 +47,6 @@ library ConstantProductOracleMath {
         uint256 sqrtK = FixedPointMathLib.sqrt(reserve0 * reserve1);
 
         int256 rawResult = LogCompression.toLowResLog(sqrtK);
-        assert(rawResult >= type(int112).min && rawResult <= type(int112).max);
```
xenide, line 51: as above

src/asset-management/AaveManager.sol
```diff
@@ -214,7 +214,7 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
         IAssetManagedPair lPair = IAssetManagedPair(msg.sender);
         int256 lAmount0Change = -int256(aToken0 ? aAmount : 0);
         int256 lAmount1Change = -int256(aToken0 ? 0 : aAmount);
-        assert(lAmount0Change < 0 || lAmount1Change < 0);
+        require(lAmount0Change < 0 || lAmount1Change < 0, "AM: RETURN_ASSET_BOTH_TOKENS");
```
OliverNChalk: Is this a require or just sanity checking during dev? Like how can this ever trigger?

xenide: with the current ternary logic we have right above, this should never trigger.

I think that this is a good sanity check to have since the pair only requests for asset returns one at a time. So this makes sure that the adjustManagement is only called for one asset

OliverNChalk: But its quite literally impossible to trigger right? There exists no way/call to trigger this require?

require(amount > 0);
lAmountChange = -amount;
require(lAmountChange < 0);

should be true for all inputs?

xenide: yup, I think with the symbolic execution power of our brains we can conclude that this would indeed never trigger. Removed the `require` statement

OliverNChalk: Haha, I tried pyrometer but couldnt get it to work on our contracts

OliverNChalk (approved):

PR #181, opened 2023-04-05
feat: reservoir timelock

### LINKED ISSUE #175: feat: set asset manager's owner behind a timelock
Make sure the reservoir deployer sets up the asset manager owner behind a timelock

xenide:
## Solution

- in terms of the default delay, I was thinking of 3 days? Then we don't need a constructor to set manually. It's changeable later anyway.


OliverNChalk:
3 days to steal all user deposits feels a little quick, what do you think about 7?

xenide:
> 3 days to steal all user deposits feels a little quick, what do you think about 7?

am totally okay with 7 days. Changed.

src/ReservoirTimelock.sol
```diff
@@ -0,0 +1,31 @@
...
+import { ReservoirPair } from "src/ReservoirPair.sol";
+import { StablePair } from "src/curve/stable/StablePair.sol";
+
+contract ReservoirTimelock is CompTimelock(msg.sender, 7 days) {
+    function setCustomSwapFee(GenericFactory aFactory, address aPair, uint256 aSwapFee) external {
+        require(msg.sender == admin, "RT: ADMIN");
+
+        bytes memory lCalldata = abi.encodeCall(ReservoirPair.setCustomSwapFee, (aSwapFee));
+        aFactory.rawCall(aPair, lCalldata, 0);
+    }
```
OliverNChalk, line 16: I was thinking this might be risky but i checked and we cap the fee at 2%, which feels fine. So I think we're all good

OliverNChalk (approved):

PR #186, opened 2023-04-10
Oliver review

OliverNChalk:


src/GenericFactory.sol
```diff
@@ -16,7 +16,7 @@ uint256 constant MAX_SSTORE_SIZE = 0x6000 - 1;
 contract GenericFactory is IGenericFactory, Owned {
     using Bytes32Lib for address;
 
-    address public stableMintBurn;
+    address public immutable stableMintBurn;
```
OliverNChalk: I was thinking this wouldnt work because ResrvoirPair stores factory as immutable. So a dual immutable reference whichis impossible

xenide: to the ReservoirPair, factory is immutable (it's stored in code), but to get stableMintBurn, it has to do an external call to factory. Pair doesn't care how factory implements stableMintBurn, whether it's immutable or not

xenide (approved):
good to go

PR #187, opened 2023-04-10
test: asset manager corner cases

xenide:
## Motivation

- in our current implementation, because the asset manager tracks holdings by shares, but the pair tracks the managed amounts by tokens, this creates a discrepancy
- in normal operating situations, this would sometimes result in balances sometimes being off by 1 
   - this means that sometimes the pairs will report a loss of 1 
   - I don't think this is something we can totally avoid, unless we keep each account's tokens in a separate address like we considered last time
   - I reviewed the `ERC4626` code as well and it seems that they circumvent this problem by having a `maxRedeem` or `maxWithdraw`. But without those, I think it's also prone to having those dust amounts leftovers
- in full redeem situations, this would result in there being a few wei of leftovers 
   - after attempting a full redeem, sometimes the pair will still have `token0/1Managed` (up to 3 in the fuzz tests) so what we may have to allow for change in manager within this range, if we don't implement the `maxRedeem` option

## Solution

- regarding `_decreaseShares` 
   - we can do `divWadUp`, then we need the check when decreasing shares to make sure it doesn't underflow 
   - earlier parties' pies are smaller when they exit. The last pair to exit will have a slightly greater share of the pie 
- potential solution
   - we introduce a `maxRedeem` that empties out by calculating the SHARES of the pair, not by underlying, for the purposes of preparing for migration. That will clean out the pair's tokens cleanly I think
   - for normal operations, we accept the occassional off-by-1  
   - we still only want the pair to know about the number of tokens, not shares, correct? Cuz the concept of shares are not meaningful to the pair's operations

### DIFF
--- a/src/asset-management/AaveManager.sol
+++ b/src/asset-management/AaveManager.sol
@@ -131,7 +131,9 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
         private
         returns (uint256 rShares)
     {
+        // note about error of `1e-decimals`
         rShares = aAmount.divWad(_getExchangeRate(aAaveToken));
+        // might want to do: if rShares are greater than the shares we have, just zero it
         shares[aPair][aToken] -= rShares;
         totalShares[aAaveToken] -= rShares;
     }
--- a/src/asset-management/AaveManager.sol
+++ b/src/asset-management/AaveManager.sol
@@ -13,7 +13,7 @@ import { IPoolAddressesProvider } from "src/interfaces/aave/IPoolAddressesProvid
 import { IPool } from "src/interfaces/aave/IPool.sol";
 import { IAaveProtocolDataProvider } from "src/interfaces/aave/IAaveProtocolDataProvider.sol";
 import { IRewardsController } from "src/interfaces/aave/IRewardsController.sol";
-import "forge-std/console.sol";
+
 contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
     using FixedPointMathLib for uint256;
 
@@ -116,7 +116,6 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
         require(lAaveUnderlying != 0, "AM: INVALID_AAVE_BALANCE");
 
         rExchangeRate = lAaveUnderlying.divWad(lTotalShares);
-        console.log("exchange rate", rExchangeRate);
     }
 
     function _increaseShares(IAssetManagedPair aPair, ERC20 aToken, ERC20 aAaveToken, uint256 aAmount)
@@ -126,7 +125,6 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
         rShares = aAmount.divWad(_getExchangeRate(aAaveToken));
         shares[aPair][aToken] += rShares;
         totalShares[aAaveToken] += rShares;
-        console.log("total shares after increase", totalShares[aAaveToken]);
     }
 
     function _decreaseShares(IAssetManagedPair aPair, ERC20 aToken, ERC20 aAaveToken, uint256 aAmount)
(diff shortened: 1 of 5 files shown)


src/asset-management/AaveManager.sol
```diff
@@ -28,8 +28,10 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
 
     /// @dev percentage of the pool's assets, above and below which
     /// the manager will divest the shortfall and invest the excess
-    uint256 public upperThreshold = 70;
-    uint256 public lowerThreshold = 30;
+    /// 1e18 == 100%
+    // TODO: can we use uint128 for each of these for gas savings?
```
xenide: would like your opinion on this @OliverNChalk

src/asset-management/AaveManager.sol
```diff
@@ -129,9 +131,18 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
+            shares[aPair][aToken] -= rShares;
+        } else {
+            // clear it out
+            shares[aPair][aToken] = 0;
+        }
+        if (totalShares[aAaveToken] >= rShares) {
+            totalShares[aAaveToken] -= rShares;
+        } else {
+            totalShares[aAaveToken] = 0;
+        }
```
OliverNChalk: Imagine scenario:

```txt
PAIR_A: 100 SHARES
PAIR_B: 100 SHARES

PAIR_A DOES FULL REDEEM
  - rShares = 101
  - shares[PAIR_A][TOKEN_A] -= 100 shares
  - totalSahres[PAIR_A][TOKEN_A] -= 101 shares
  
PAIR_B DOES FULL REDEEM
  - rShares = 101
  - shares[PAIR_B][TOKEN_A] -= 100 shares
  - totalSahres[PAIR_B][TOKEN_A] -= 99 shares
```

This does not result in locked funds but leads to the invariant `sum(balances) == totalSupply` breaking. I propose we instead do the following:
```solidity
        rShares = aAmount.divWadUp(_getExchangeRate(aAaveToken));
        if (rShares >= shares[aPair][aToken]) { rShares = shares[aPair][aToken]; }

        shares[aPair][aToken] -= rShares;
        totalShares[aAaveToken] -= rShares;
```

src/asset-management/AaveManager.sol
OliverNChalk: Yeah, this should save gas and I'm happy with changing

OliverNChalk (changes requested):

src/asset-management/AaveManager.sol
xenide: okay never mind, changing it actually INCREASED gas usage across the board so leaving it for now

OliverNChalk: are you sure, did you use forge snapshot --diff.

do you deploy AaveManager in setUp or in test cases? Make sure you're benchmarking runtime cost not deployment cost

xenide: I didn't use the diff, generated two files and compared manually. 

Because of the way we're using the modifiers for multiple networks + pairs, we can't instantiate the manager in setUp but instead have to do it in a helper function. So in our case there's no way to completely separate out the deployment cost. Unless we create a new test file

src/asset-management/AaveManager.sol
xenide: good thinking. Implemented. The only thing I want to highlight is that the condition should be strictly greater, not greater or equal to 

`rShares > shares[aPair][aToken]`

src/asset-management/AaveManager.sol
OliverNChalk: Okay, so your gas test does not conclude anything then. Happy to leave as uint256 as we can always iterate

src/asset-management/AaveManager.sol
OliverNChalk: good catch, you're right

OliverNChalk (approved):

src/asset-management/AaveManager.sol
xenide: for record purposes, we have use the `gasleft()` method to show that it indeed decreases gas usage and thus we've implemented it

PR #191, opened 2023-05-06
fix: docs for audit 

xenide:


src/GenericFactory.sol
```diff
@@ -130,6 +130,7 @@ contract GenericFactory is IGenericFactory, Owned {
 
     event Pair(address indexed token0, address indexed token1, uint256 curveId, address pair);
 
+    // maps tokenA, tokenB addresses to pair address, where the order of tokenA and tokenB does not matter
```
OliverNChalk: Should be natspec?

src/curve/stable/StablePair.sol
```diff
@@ -302,6 +294,9 @@ contract StablePair is ReservoirPair {
         int112 currLogLiq = ConstantProductOracleMath.calcLogLiq(aReserve0, aReserve1);
         prevClampedPrice = currClampedPrice;
 
+        // overflow is desired here as the consumer of the oracle will be reading the difference in those
+        // accumulated log values
+        // when the index overflows it will overwrite the oldest observation and then forms a loop
```
OliverNChalk, line 297: Should mention the case where we have no swaps for N years and the timestamp overflows twice

src/interfaces/IReservoirCallee.sol
```diff
@@ -2,7 +2,9 @@
 pragma solidity ^0.8.0;
 
 interface IReservoirCallee {
+    /// @param sender the target address for the arbitrary data call in the case of a flash swap
```
OliverNChalk:
```suggestion
    /// @param user that initiated the swap and is triggering the callback
```

src/interfaces/IReservoirCallee.sol
```diff
@@ -2,7 +2,9 @@
...
     /// @param amount0 positive indicates the amount out (received by callee), negative indicates the amount in (owed by callee)
     /// @param amount1 positive indicates the amount out (received by callee), negative indicates the amount in (owed by callee)
+    /// @param data the bytes that are provided in the case of a flash swap
```
OliverNChalk:
```suggestion
    /// @param data provided by the user is returned as part of the callback
```

src/asset-management/AaveManager.sol
```diff
@@ -101,16 +101,12 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
-
-    function setLowerThreshold(uint128 aLowerThreshold) external onlyOwner {
-        require(aLowerThreshold <= 1e18 && aLowerThreshold <= upperThreshold, "AM: INVALID_THRESHOLD");
+    function setThresholds(uint128 aLowerThreshold, uint128 aUpperThreshold) external onlyOwner {
+        require(aUpperThreshold <= 1e18 && aUpperThreshold >= aLowerThreshold, "AM: INVALID_THRESHOLDS");
         lowerThreshold = aLowerThreshold;
+        upperThreshold = aUpperThreshold;
         emit LowerThreshold(aLowerThreshold);
+        emit UpperThreshold(aUpperThreshold);
     }
```
OliverNChalk: Should we not collpase the logs into a single one?

xenide: good thinking, yes we should. Fixed

src/interfaces/IReservoirCallee.sol
xenide: fixed

src/interfaces/IReservoirCallee.sol
xenide: fixed

src/GenericFactory.sol
xenide: fixed

src/curve/stable/StablePair.sol
xenide: okay added the comment, but added it in ReservoirPair L147 as I think it's a more appropriate venue for this comment

OliverNChalk (approved):

PR #192, opened 2023-05-06
fix: audit logic changes 

xenide:


src/asset-management/AaveManager.sol
```diff
@@ -178,7 +170,7 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
             return 0;
         }
 
-        rTokenBalance = shares[aOwner][aToken].mulWad(_getExchangeRate(lAaveToken));
+        rTokenBalance = shares[aOwner][aToken] * lAaveToken.balanceOf(address(this)) / lTotalShares;
```
OliverNChalk, line 168: Do we currently have the issue where pools will report a 1 wei loss if on adjustManagement/similar?

src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -86,8 +86,6 @@ contract ConstantProductPair is ReservoirPair {
                     }
                 }
             }
-        } else if (kLast != 0) {
-            kLast = 0;
```
OliverNChalk, line 92: Interesting, this was actually a bug but they didn't flag it as such. We should keep this change but also remove the return value `bool rFeeOn`.

Additionally, I think we can remove `if (lSqrtOldK != 0)` but lets do that later so we can add an issue now tracking it

OliverNChalk: This is also the case for Stablepair i believe

OliverNChalk (approved):
I think the klast != 0 / lastInvariant != 0 changes are bugs

src/asset-management/AaveManager.sol
xenide: for small amounts and if there are multiple `adjustManagement` + `getBalance` calls within one block, then yes. 

But as soon as some time pass and some interest accumulates there this "small" loss would not happen

OliverNChalk: Okay, probably not much we can do unless we wanted to have some sort of deadband. Happy to leave as is

src/curve/constant-product/ConstantProductPair.sol
xenide: alright so after reading your comment, I thought we still needed this check because in `_calcFee`, it's dividing this by this number so if we remove the check we're going to have a division by zero error. E.g. for first mints. 

But I tested it out and it doesn't happen in both curves, due to the `lSqrtNewK > lSqrtOldK` check. In the first mint everything is zero so this problem wouldn't be triggered. 

Created an issue to track this

OliverNChalk (approved):

PR #195, opened 2023-05-09
fix: audit casting issues 

xenide:


src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -58,9 +58,12 @@ contract ConstantProductPair is ReservoirPair {
...
     {
-        // ASSERT: newK & oldK        < uint104
-        // ASSERT: aPlatformFee       < FEE_ACCURACY
-        // ASSERT: aCirculatingShares < uint104
+        // INVARIANT: aSqrtOldK & aSqrtNewK < uint104 as _syncManaged ensures that both reserves fit into uint104.
+        //            The sqrt of the product of the two reserves will fit into uint104 as well
+        // INVARIANT: aSqrtOldK < aSqrtNewK as checked in _mintFee
+        // INVARIANT: aPlatformFee       < FEE_ACCURACY
+        // INVARIANT: aCirculatingShares < uint104  since the circulating shares are the geometric mean of the reserves
+        //            and that both reserves fit into uint104 as explained above, aCirculatingShares will fit into uint104 as well
```
OliverNChalk:
```suggestion
        // INVARIANT: `aSqrtOldK & aSqrtNewK < uint104` as _syncManaged ensures that both reserves fit into uint104.
        //            The sqrt of the product of the two reserves will fit into uint104 as well
        // INVARIANT: `aSqrtOldK < aSqrtNewK` as checked in _mintFee
        // INVARIANT: `aPlatformFee < FEE_ACCURACY` enforced by setter functions
        // INVARIANT: `aCirculatingShares < uint104`  since the circulating shares are the geometric mean of the reserves
        //            and that both reserves fit into uint104 as explained above, aCirculatingShares will fit into uint104 as well
```

OliverNChalk: Also, is the last invariant true? Could it be broken if the pool's initial exchange rate is vastly different to its current?

My gut tells me this probably holds because as the pools composition changes the exchange rate will compensate to prevent the shares getting out of hand.

What's your thought on this?

OliverNChalk (commented):
Looks good, just 1 minor comment re the invaraints

src/curve/constant-product/ConstantProductPair.sol
xenide: Alright so to test out this hypothesis, I wrote several test cases: 

1. a CP starting with perfectly balanced tokens
    - adding liq to the max (both reserves at `uint104.max`) => total supply of LP tokens also at `uint104.max`
   - swap to change the ratio to be very skewed, then adding liq to the max => total supply far below uint104.max

2. a CP starting with very skewed ratios
   - adding liq to the max => total supply far below uint104.max
   - swap to change the ratio to be roughly balanced, then add liq to the max => total supply is near `uint104.max`

So I think our hypothesis holds. The only scenario where the LP token supply can even reach uint104 is when both tokens are perfectly balanced and their reserves reach uint104. Even in that case it will not exceed uint104. 

So I think we can resolve this issue.

OliverNChalk (approved):

src/curve/constant-product/ConstantProductPair.sol
OliverNChalk: Agree

PR #196, opened 2023-05-10
fix: make virtual getters public and actual variables private

xenide:
## Solution

seems that the gas performance actually... increased?


OliverNChalk (approved):
Approved assuming there is no runtime gas penalty in doing it this way?

xenide:
> Approved assuming there is no runtime gas penalty in doing it this way?

if you look at the gas tests it actually saved about 3 gas doing it this new way

PR #199, opened 2023-05-14
fix: phantom overflow

xenide:


src/curve/stable/StableMintBurn.sol
```diff
@@ -110,8 +120,8 @@ contract StableMintBurn is StablePair {
             lTotalSupply = totalSupply;
         }
 
-        rAmount0 = liquidity * lReserve0 / lTotalSupply;
-        rAmount1 = liquidity * lReserve1 / lTotalSupply;
+        rAmount0 = liquidity.mulDiv(lReserve0, lTotalSupply);
+        rAmount1 = liquidity.mulDiv(lReserve1, lTotalSupply);
```
xenide: decided to fix this as there is no guarantee that liquidity (on the same scale as `totalSupply`) will fit within 256-104=152 bits. Also recorded the reasoning in the spreadsheet

src/ReservoirPair.sol
```diff
@@ -512,13 +514,18 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
             return (aCurrRawPrice, int112(LogCompression.toLowResLog(aCurrRawPrice)));
         }
 
+        // call to `percentDelta` is safe as long as the difference between aCurrRawPrice and aPrevClampedPrice is
+        // less than uint196 (1e59). It is extremely unlikely that one trade can change the price by 1e59
         if (aCurrRawPrice.percentDelta(aPrevClampedPrice) > maxChangeRate * aTimeElapsed) {
```
xenide: on a technical level, it is possible to overflow this with a very large and a very small number. But I think it's so unlikely that I'm willing to let this one swap revert if this case happens

src/ReservoirPair.sol
```diff
@@ -512,13 +514,18 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
             // clamp the price
+            // multiplication of maxChangeRate and aTimeElapsed would not overflow as
+            // maxChangeRate <= 0.01e18 (50 bits)
+            // aTimeElapsed <= 32 bits
             if (aCurrRawPrice > aPrevClampedPrice) {
-                rClampedPrice = aPrevClampedPrice * (1e18 + (maxChangeRate * aTimeElapsed)) / 1e18;
+                rClampedPrice = aPrevClampedPrice.mulDiv(1e18 + maxChangeRate * aTimeElapsed, 1e18);
```
xenide, line 526: max price the oracle can record is `e^130` which is around 248 bits. So phantom overflow is possible in this case, thus the fix

src/libraries/ConstantProductMath.sol
```diff
@@ -4,6 +4,12 @@ pragma solidity ^0.8.0;
 library ConstantProductMath {
     uint256 public constant FEE_ACCURACY = 1_000_000; // 100%
 
+    /// @dev the function assumes that the following args are within the respective bounds as enforced by ReservoirPair
+    /// and therefore would not overflow
+    /// aAmountIn   <= uint104
+    /// aReserveIn  <= uint104
+    /// aReserveOut <= uint104
+    /// aSwapFee    <= 0.2e6
     function getAmountOut(uint256 aAmountIn, uint256 aReserveIn, uint256 aReserveOut, uint256 aSwapFee)
```
OliverNChalk: You should mention that breaching these args is safe (it will revert) rather than invalid behaviour

src/libraries/StableMath.sol
```diff
@@ -64,6 +66,8 @@ library StableMath {
         uint256 swapFee,
         uint256 N_A // solhint-disable-line var-name-mixedcase
     ) internal pure returns (uint256 dx) {
+        // overflow and underflow are not possible as reserves, amountIn <= uint104 and precision multipliers are maximum 1e18 (uint60)
+        // and swapFee < ONE_HUNDRED_PERCENT
```
OliverNChalk, line 70: Do we check all callsites?

src/ReservoirPair.sol
OliverNChalk: You need to say that if this happens it will revert. You will make people think we can be hacked with this comment

src/ReservoirPair.sol
OliverNChalk: Makes sense

src/ReservoirPair.sol
xenide: updated comment to say that it will revert when the inputs are beyond a certain range

src/libraries/ConstantProductMath.sol
xenide: fixed

src/libraries/StableMath.sol
xenide: just did a check. 

Seems that for `_getAmountOut` (exact in trades), it is indeed possible for the input amountIn to be more than `uint104`. But this would be caught at `_updateAndUnlock` or the overflow in `_computeLiquidityFromAdjustedBalances`. So I think there's no need to do anything. If we want to be super safe we can add a check `require(lAmountIn < type(uint104).max)` in StablePair. Currently a similar check exist for exact out trades. 

Also reviewed the analogous cases in ConstantProduct. It's not a concern for ConstantProduct even if the amountIn exceeds the range cuz the math is checked (so it reverts when overflow), whereas the math is unchecked for Stable (for `_getAmountIn` and `_getAmountOut`. The iterative approx functions are checked).

OliverNChalk (approved):

PR #200, opened 2023-05-15
fix: bring back try catch for burn 

xenide:


src/curve/stable/StableMintBurn.sol
```diff
@@ -127,6 +132,11 @@ contract StableMintBurn is StablePair {
         revert("SMB: IMPOSSIBLE");
     }
 
+    function mintFee(uint256 aReserve0, uint256 aReserve1) external virtual override returns (uint256, uint256) {
```
OliverNChalk: Name return values

OliverNChalk (approved):
Looks good, nice work

src/curve/stable/StableMintBurn.sol
xenide: fixed

PR #201, opened 2023-05-17
fix(generic-factory): use yul loop to ensure memory safety

OliverNChalk:


xenide (approved):
logic looks correct to me

PR #205, opened 2023-05-23
test: differential test solady mulDiv with solmate and oz

xenide:


xenide:
personally I don't think there's a need to move these into the "differential" tests family, as opposed to leaving it in the unit tests. Just that they will have an entry in the gas snapshot

OliverNChalk (approved):

PR #210, opened 2023-06-13
fix: rm redundant check for platform fee invariant

### LINKED ISSUE #193: fix: remove `lSqrtOldK != 0` check for ConstantProduct and `lDLast != 0` check for Stable

xenide:


src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -77,18 +77,15 @@ contract ConstantProductPair is ReservoirPair {
 
         if (lFeeOn) {
             uint256 lSqrtOldK = FixedPointMathLib.sqrt(kLast); // gas savings
+            uint256 lSqrtNewK = FixedPointMathLib.sqrt(aReserve0 * aReserve1);
 
-            if (lSqrtOldK != 0) {
-                uint256 lSqrtNewK = FixedPointMathLib.sqrt(aReserve0 * aReserve1);
+            if (lSqrtNewK > lSqrtOldK) {
```
OliverNChalk, line 82: What happens in the scenario where we turn platform fee off? Won't platform fee be charged on the full range because sqrt_old_k is zero?

xenide: `kLast` is always written to at the end of mint+burn regardless if platform fee is on or off. 

There is a check of `feeOn` at the beginning of the function. So if platform fee was off, this wouldn't even happen. So the scenario you described will not take place

xenide: If we consider more extreme scenarios, such as turning platform fee off and then on again, it would also work properly since the last invariant is always written to

OliverNChalk: sounds good

OliverNChalk (approved):

PR #219, opened 2023-09-29
test: attempting to redeem while AAVE fully borrowed

xenide:
## Motivation

- in this PR / issue I want to explore the scenario where AAVE's market is fully borrowed (utilization at 100%) and the pair tries to redeem, as mentioned by Michael from Euler previously

## Solution

- so this can occur when: 
   - some guy with a lot of collateral borrows the entire available liq
   - the pair receives a swap / burn request that demands more than its liq balance (30% as set by the manager) in the pair
- implications: 
  - our oracle cannot be updated if the swap or burn fails, resulting in inaccurate prices for as long as we cannot redeem
- mitigated by: 
  - attacker has to endure the sky high interest rates when borrowing the entirety of the liquidity
  - market will come provide liq when interest rates are so high

- given the conditions to make it happen and the ongoing costs, it seems that this attack is highly unlikely to be pulled off, and if did, would not be sustainable even for a few hours

- on hindsight I didn't even need to write a test case to show that the redemption would have failed, common sense would have sufficed

### DIFF
--- a/src/interfaces/aave/IPool.sol
+++ b/src/interfaces/aave/IPool.sol
@@ -77,4 +77,22 @@ interface IPool {
      *
      */
     function withdraw(address asset, uint256 amount, address to) external returns (uint256);
+
+    /**
+     * @notice Allows users to borrow a specific `amount` of the reserve underlying asset, provided that the borrower
+     * already supplied enough collateral, or he was given enough allowance by a credit delegator on the
+     * corresponding debt token (StableDebtToken or VariableDebtToken)
+     * - E.g. User borrows 100 USDC passing as `onBehalfOf` his own address, receiving the 100 USDC in his wallet
+     *   and 100 stable/variable debt tokens, depending on the `interestRateMode`
+     * @param asset The address of the underlying asset to borrow
+     * @param amount The amount to be borrowed
+     * @param interestRateMode The interest rate mode at which the user wants to borrow: 1 for Stable, 2 for Variable
+     * @param referralCode The code used to register the integrator originating the operation, for potential rewards.
+     *   0 if the action is executed directly by the user, without any middle-man
+     * @param onBehalfOf The address of the user who will receive the debt. Should be the address of the borrower itself
+     * calling the function if he wants to borrow against his own collateral, or the address of the credit delegator
+     * if he has been given credit delegation allowance
+     */
+    function borrow(address asset, uint256 amount, uint256 interestRateMode, uint16 referralCode, address onBehalfOf)
+        external;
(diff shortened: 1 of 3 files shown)


OliverNChalk:
To make this scenario concrete:

EXMP token price capped to the upside:
- Attacker targets EXMP<>WETH pool.
- EXMP<>WETH puts 50% of reserves in Aave.
- Attacker borrows all EXMP from Aave, preventing the pool from withdrawing more.
- Attacker then buys remaining EXMP in Reservoir pool, pushing the price up.
- The oracle can now no longer show higher prices than the current for EXMP.
- Is this a risk, users can still sell EXMP down so the price is only stuck from going up.
- Seems like this is not a risk unless EXMP is genuinely mooning and there is high borrow interest during the jump in price.

EXMP token price capped to the downside:
- Attacker targets EXMP<>ILQD pool.
- EXMP<>ILQD puts 50% reserves in Aave.
- Attacker borrows all ILQD from Aave.
- Attacker buys all ILQD from pool (by dumping EXMP).
- Price is now capped to the downside and cannot go down further.
- If ILQD is genuinely cratering with high borrow interest, this could be a problem as this would leave bad debt in any MM on top of Reservoir.,

In my mind the main risk here:

- Do not pair project tokens with high risk assets (focus on blue chips as quote asset).
- It's unlikely that a project token jumps a large % quickly without time for arbitrageurs to provide more liquidity to the market.
- Even so, project tokens should still be sufficiently liquid if they are going to be borrowed (else in event of a massive price jump users could borrow the project token at a stale price).


Mitigation wise, there may be a way to detect the pools inability to withdraw and thus disable further borrowing while there is high interest.

xenide:
Coming back to this problem, I agree with using blue chip assets as quote assets. 

For the mitigation, we can use a variable in the Vault / Risk manager, to track how much assets in the pool belongs to the AMM pairs. E.g. pool size is 10M, and 3M of which belongs to the pairs, the last 3M cannot be borrowed out, to ensure that the pairs can always redeem (to fulfill swap / burn obligations). This can be implemented as a vault status check (which is done at every borrow anyway, usually validating supply and borrow caps) and seems doable to me. Of course we need to be careful with the implementation such that in the event of a loss, it gets socialized to these pairs' assets too so that the suppliers don't get overly penalized. Agree?

xenide:
After discussion, we determined that we don't need to build in additional safeguards as: 

> Additionally an arbitrageur could atomically supply USDC and swap to capture the arb + receive Apr until the supply imbalance fixes

Closing for now

PR #220, opened 2024-01-22
fix: oracle written timestamp

xenide:
## Motivation

- Oracle observations were incorrectly written with the timestamp of the last activity block, instead of the current activity block. This PR fixes the mistake by writing `block.timestamp` (the 31 bit version of it) as the observation's timestamp

## Solution

- Integration tests breaking due to a foundry bug, [tweet](https://twitter.com/programmersmart/status/1744515678349213908?s=61&t=7CdQOlkVOSluPSUoJYQe4g). Not fixing for now
- For additional tests, there aren't any except for one that checks that the latest observation indeed stores the latest block timestamp. The previous tests are still valid, and because they use relative time, this bug didn't surface. In a way, this bug is only easier to spot when we are trying to query the oracle beyond the latest observation


test/unit/AssetManagedPair.t.sol
```diff
@@ -208,8 +208,8 @@ contract AssetManagedPairTest is BaseTest {
 
     function testMint_AfterLoss(uint256 aNewManagedBalance0, uint256 aNewManagedBalance1) external allPairs {
         // assume
-        uint256 lNewManagedBalance0 = bound(aNewManagedBalance0, 1, 10e18);
-        uint256 lNewManagedBalance1 = bound(aNewManagedBalance1, 1, 10e18);
+        uint256 lNewManagedBalance0 = bound(aNewManagedBalance0, 1, 9e18);
+        uint256 lNewManagedBalance1 = bound(aNewManagedBalance1, 1, 9e18);
```
OliverNChalk, line 213: What was the reason fro 10 -> 9?

xenide: Reason for this is: due to the update for foundry the fuzzes are getting more coverage at the extreme values. So sometimes if it gets to `10e18` then there is no loss for the pair, and the redeemer does not get back fewer tokens, making the assert condition below fail. 

I lowered the range to `9e18` to make sure there's a loss for the pair, in line with the spirit of the test. Else the alternative would be to change the assert condition from `assertLt` to `assertLte`, which would be smellier code.

OliverNChalk: Makes sense

OliverNChalk (approved):

PR #221, opened 2024-01-22
feat: use mapping instead of array for oracle observations

xenide:
## Motivation

- inspiration came from balancer as they use a mapping to simulate an array
- UniV3 still uses an array of size 65535 (one less than uint16.max)
- since our use case is the same, always knowing the index, I thought it'd be more efficient to use a mapping over an array 
- Arrays would make sense if each element is smaller than a word 

## Solution
 
- it is indeed more gas efficient. Will upload the snapshot after the previous PR is merged 
- also tested with `uint16` as the input type instead of `uint256`. But the gas changes are a mixed bag. Therefore not proceeding with that change


OliverNChalk (approved):
approved assuming it's cheaper

test/unit/StablePair.t.sol
```diff
@@ -924,7 +924,7 @@ contract StablePairTest is BaseTest {
         bytes32 lEncoded = bytes32(abi.encodePacked(lLastInvariantAmp, lLastInvariant));
         // hardcoding the slot for now as there is no way to access it publicly
         // this will break when we change the storage layout
-        vm.store(address(_stablePair), bytes32(uint256(65_553)), lEncoded);
+        vm.store(address(_stablePair), bytes32(uint256(18)), lEncoded);
```
xenide, line 927: had to change the storage slot as the storage layout has changed cuz we're no longer using an array for `_observations`

PR #222, opened 2024-02-06
fix: `returnAsset` exploit

xenide:
## Motivation

- bug was due to 
```solidity
uint256 lCurrentShares = shares[aPair][aToken];
// this is to prevent underflow as we round up in the previous division operation
if (rShares > lCurrentShares) {
   rShares = lCurrentShares;
}
```

- code was introduced in #187 

- `mulDivUp` behavior
  - does not round up if division is exact, only rounds up when there is remainder

- in our current production deployment of asset manager, all pairs have redeemed successfully (tokens managed are all 0 from the perspective of the pair), but individual and total shares are more than 0. Of course we have a special case where there was a loss (wasn't strictly increasing)

Screenshot 2024-02-22 at 10 29 07

## Potential Solutions

1. add a condition to check that `lCurrentShares` is greater than 0

```solidity
if (rShares > lCurrentShares && lCurrentShares > 0) 
``` 
- this is not secure, as `lCurrentShares` can be 1, and `rShares` can be a really large number. i.e. an attacker can get several shares by letting the asset manager manage its assets, and then still attack using `returnAsset` 

2. remove `rShares > lCurrentShares` check completely, relying on checked arithmetic. 

- this is the recommended solution (implemented in this PR), as underflow will occur in the case of attempting to redeem 0 shares
- I've attempted to break this by fuzzing 100K times. So far none of them have triggered the subtraction underflow problem
  - the total number of shares and individual number of shares may not end up being 0, but at least the `token0/1Managed` will be zero. 
    - this is only happening for very small amounts (e.g. amounts less than 10 wei)
- are there specific numbers to break it? 

3. any other solutions you can think of?

### DIFF
--- a/test/integration/Aave.t.sol
+++ b/test/integration/Aave.t.sol
@@ -534,8 +534,8 @@ contract AaveIntegrationTest is BaseTest {
         // assume
         (uint256 lReserve0, uint256 lReserve1,,) = _pair.getReserves();
         uint256 lReserveUSDC = _pair.token0() == USDC ? lReserve0 : lReserve1;
-        int256 lAmountToManage1 = int256(bound(aAmountToManage1, 100, lReserveUSDC / 2));
-        int256 lAmountToManage2 = int256(bound(aAmountToManage2, 100, lReserveUSDC / 2));
+        int256 lAmountToManage1 = int256(bound(aAmountToManage1, 1e6, lReserveUSDC / 2));
+        int256 lAmountToManage2 = int256(bound(aAmountToManage2, 1e6, lReserveUSDC / 2));
 
         // arrange
         (address lRawAaveToken,,) = _dataProvider.getReserveTokensAddresses(address(USDC));
--- a/test/integration/Aave.t.sol
+++ b/test/integration/Aave.t.sol
@@ -145,7 +145,6 @@ contract AaveIntegrationTest is BaseTest {
 
     function setUp() external {
         _networks.push(Network(getChain("avalanche").rpcUrl, 0xB97EF9Ef8734C71904D8002F8b6Bc66Dd9c48a6E, 0xB7887FED5E2f9dc1A66fBb65f76BA3731d82341A));
-//        _networks.push(Network(getChain("polygon").rpcUrl, 0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174, ));
 
         vm.makePersistent(address(_tokenA));
         vm.makePersistent(address(_tokenB));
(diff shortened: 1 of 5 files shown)


OliverNChalk:
This change was originally introduced to ensure a full redemption was possible right? I think provided we have a way to recover if this maths breaks in full redemption mode then we can be reasonably comfortable. I.e. it may make sense to give the admin super user style power to make calls from the asset manager if needed?

You could imagine a function like this:
```solidity
function rawCall(address target, uint256 value, bytes data) external onlyOwner
```

xenide:
So previously when we did the PR, we were relying on an `_getExchangeRate` internal function, which was only accurate to 18 decimal places, which probably resulted in a few wei of difference. 

Then during the audit, the auditors recommended us to perform the arithmetic directly (without using an intermediate exchange rate) which probably eliminated this problem. That said, with the testing I've done, I haven't been able to come up with a scenario in which full redemption will fail. 

Yes I thought of the same regarding introducing `rawCall` too. If we put it behind a timelock I think it's acceptable to everyone. Will implement

OliverNChalk:
> So previously when we did the PR, we were relying on an `_getExchangeRate` internal function, which was only accurate to 18 decimal places, which probably resulted in a few wei of difference.
> 
> Then during the audit, the auditors recommended us to perform the arithmetic directly (without using an intermediate exchange rate) which probably eliminated this problem. That said, with the testing I've done, I haven't been able to come up with a scenario in which full redemption will fail.
> 
> Yes I thought of the same regarding introducing `rawCall` too. If we put it behind a timelock I think it's acceptable to everyone. Will implement

Assuming the owner is the timelock, it should be acceptable?

test/integration/Aave.t.sol
```diff
@@ -117,25 +122,36 @@ contract AaveIntegrationTest is BaseTest {
...
+            vm.startPrank(masterMinterUSDC);
+            IUSDC(address(USDC)).configureMinter(masterMinterUSDC, type(uint256).max);
+            IUSDC(address(USDC)).mint(aRecipient, aAmount);
+            vm.stopPrank();
+        }
+    }
+
+    function setUp() external {
+        _networks.push(Network(getChain("avalanche").rpcUrl, 0xB97EF9Ef8734C71904D8002F8b6Bc66Dd9c48a6E, 0xB7887FED5E2f9dc1A66fBb65f76BA3731d82341A));
+//        _networks.push(Network(getChain("polygon").rpcUrl, 0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174, ));
```
OliverNChalk: Remove?

OliverNChalk (approved):

test/integration/Aave.t.sol
xenide: rm-ed

PR #223, opened 2024-03-04
fix: write instant price to oracle and don't write liquidity data

xenide:
## Motivation

- when calculating TWAP, we need the instant price to calculate the accumulator that has elapsed since the last oracle write (which is almost all the time)
- as discussed within the team, for a given base and quote token, the price feed (pair) to be used will be controlled by governance instead of which pair has more liquidity, so we're doing away with writing liquidity data in the observations 

- Additional math changes: the following problems have always been there but because the way we write tests we don't progress the time, no oracle writing was done so it wasn't caught. Now that we update the instant price at every block, we caught it
  1. `StableMath::_computeLiquidityFromAdjustedBalances` would overflow sometimes given very large inputs 
      - Solution: introduce `fullMulDiv` for iterative algorithm for 2 arithmetic operations as overflow can happen for large numbers in those specific operations. 
  2. `StableOracleMath::calcSpotPrice` would sometimes fail when normalized reserves are too small (e.g. 1e3) such that when multiplied and wad-ed down, the result is 0
      - Solution: return a default price of `1e18` if either numerator / denominator is 0 
        - Risk assessment: if we're using the pair as an oracle probably it will have more liquidity than just 1-2 tokens (in normalized terms). And if there's so few tokens (1e3) we wouldn't be using the pair as a price feed. 
      - Other solutions explored: 
        - scaling all reserves by 1e18 if any of the reserves are < 1e18. This causes overflows in some test cases
        - checking if only `derivativeY` is 0, and then setting it to 1. This is insufficient as `derivativeX` can be zero too and make the function return 0 then cause a problem with the log compression
      - Potential solutions not yet explored: 
         - instead of calculating precisely the spot price, we estimate it with simulating a swap with zero fees 0.01% of the pool's liquidity (to simulate a very small input). This should be close enough 
         - use another impl to calculate spot price. Though this impl may be susceptible to the same problem we're facing (going to 0 for small inputs)
  - max change per change  #228 
  - reducing oracle buffer size #229


src/ReservoirPair.sol
```diff
@@ -148,15 +148,45 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
+                (, int256 lLogInstantClampedPrice) = _calcClampedPrice(
+                    lInstantRawPrice,
+                    lLogInstantRawPrice,
+                    LogCompression.fromLowResLog(lPrevious.logInstantClampedPrice),
+                    lTimeElapsed,
+                    lPrevious.timestamp
+                );
+
+                _updateOracleInstantPrices(lPrevious, lLogInstantRawPrice, lLogInstantClampedPrice);
+            }
```
OliverNChalk: We have to run this every time because we want to store the latest instant price in the oracle slot right? Is it not sufficient to only update instant price with rest of oracle inputs?

xenide:
Regarding the possibility of removing the `else` branch to only update the instant price. 

Option 1: 
- we would need to store the first ever price (price during pair creation + mint) in a storage variable 
  - this storage variable only gets read once, wasteful  
-

src/ReservoirPair.sol
```diff
@@ -148,15 +148,35 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
-        if (lTimeElapsed > 0 && aReserve0 != 0 && aReserve1 != 0) {
-            _updateOracle(aReserve0, aReserve1, lTimeElapsed, lBlockTimestamp);
+
+        // both balance should never be zero, but necessary to check so we don't pass 0 values into arithmetic operations
+        // on the first activity of every block, accumulate using previous instant prices and write the current instant prices
+        // we create a new sample even at the time of pair creation
+        // this implies that we do not update instant prices for subsequent mint/burn/swaps in the same block
+        // but the team has assessed that this is a small risk given the very fast block times on L2s
+        // and has decided to make the tradeoff to minimize complexity
+        if (lTimeElapsed > 0 && aBalance0 & aBalance1 != 0) {
```
xenide: what do you think of this math trick to ensure neither balances are 0? Seems to be more gas efficient across the board based on my tests

OliverNChalk: Shouldn't it be `aBalance0 | aBalance1 != 0`? Because if either is zero then you don't want to enter the branch?

xenide: for your condition, if either balance0 or balance1 is not zero, that's sufficient for the statement to evaluate to true, which is not what we want. We want to enter the branch when BOTH are not zero

xenide: actually, I just realized that both won't work. 


### For the & case 

bal0 = 5 (0b101) 
bal1 = 2  (0b010)

both balances are above zero, but will still evaluate to false 

### For the | case 

bal0 = 10 (0b1010)
bal1 = 0 (0x0)

will evaluate to true, which is not what we want. 

I'll undo the changes

xenide: changes undone

OliverNChalk: > For the | case
> bal0 = 10 (0b1010)
> bal1 = 0 (0x0)
> 
> will evaluate to true, which is not what we want.

This is wrong, we want |. The code before did `balance0 != 0 && balance1 != 0` which is equivalent to `balance0 | balance1 != 0`

xenide: but... 
```
0b1111 | 0b0000  != 0
```

evaluates to true, no?? Am I delusional?

OliverNChalk: Ahh, didn't realize the original statement was checking both were non zero rather than one was non zero. `|` can be used to detect non zero but `&` cannot be used to check for the presence of a zero.

Original:
```solidity
lTimeElapsed > 0 && aBalance0 > 0 && aBalance1 > 0
```

Optimized:
```solidity
unchecked {
    lTimeElapsed * aBalance0 * aBalance1 > 0
}
```

xenide: implemented the unchecked multiplication, it indeed does save a bit of gas

OliverNChalk: Is it possible to have it overflow and land exactly on 0? If so what is the consequence?

xenide: spent some time analyzing. Seems that it's not possible to overflow.

`lTimeElapsed` has max value of uint32 (limited by the type itself)
`aBalance0` and `aBalance1` are max uint104 (as checked by the require statement right above) 

So it's not possible to overflow the multiplication beyond uint256

OliverNChalk: Awesome

src/ReservoirPair.sol
```diff
@@ -135,8 +135,11 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
         _writeSlot0Timestamp(sSlot0, aBlockTimestampLast, false);
     }
 
-    // update reserves with new balances
-    // on the first call per block, update price and liq oracle using previous reserves
+    /// @notice update reserves with new balances
+    /// @notice on the first call per block, accumulate price oracle using previous instant prices and write the new instant prices
+    /// @dev we write an oracle sample even at the time of pair creation. Also we do not update instant prices for
+    /// subsequent mint/burn/swaps in the same block. the team has assessed that this is a small risk given the very
+    /// fast block times on L2s and has decided to make the tradeoff to minimize complexity
```
OliverNChalk: Would typically use proper grammar for natspec (capitalize first letter). Additionally, would say that this comment does not provide much comfort as it says we are taking a known risk instead of explaining why this is fine. Would suggest wording along the lines of "The price is not updated on subsequent stops as manipulating the instaneous price does not materially affect the TWAP, especially when using the clamped pricing."

src/ReservoirPair.sol
```diff
@@ -148,15 +151,30 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
+                    lInstantRawPrice,
+                    lLogInstantRawPrice,
+                    LogCompression.fromLowResLog(lPrevious.logInstantClampedPrice),
+                    lTimeElapsed,
+                    aBlockTimestampLast // assert: aBlockTimestampLast == lPrevious.timestamp
+                );
+                _updateOracleNewSample(
+                    sSlot0, lPrevious, lLogInstantRawPrice, lLogInstantClampedPrice, lTimeElapsed, lBlockTimestamp
+                );
+            }
```
OliverNChalk: Is the change `_updateOracle` -> `_updateOracleNewSample` necessary anymore?

OliverNChalk: This is not easy to review as it is a big diff, will make auditors scared.

OliverNChalk (changes requested):

src/ReservoirPair.sol
xenide: okay updated this comment as well as other NatSpec comments introduced in this PR to have period at the beginning and to end with a period.

test/unit/ReservoirPair.t.sol
```diff
@@ -206,4 +207,35 @@ contract ReservoirPairTest is BaseTest {
...
+
+        Observation memory lObs = _oracleCaller.observation(_pair, lIndex);
+
+        console.log(LogCompression.fromLowResLog(lObs.logInstantRawPrice));
+        console.log(LogCompression.fromLowResLog(lObs.logInstantClampedPrice));
+
+        // Attacker could repeat this for every second by only paying the swap fees. Thw mitigation factors on our end are that:
+        // 1. Arbitrage bots could submit txs that get sandwiched between the two malicious txs. How realistic is this if the attacker broadcasts 2 txs simultaneously (or even better, make it atomic via a contract call)?
+        //    - seems quite unlikely to me, cuz by the time the the block is included, the price would not have changed much to the arb bots, unless they are really sophisticated
+        // 2. The effect of the clamp
```
OliverNChalk: What does this attack achieve? Every time the attacker trades back to normal price they update the accumulator with the correct price. So they're actually just making the oracle more accurate? The instant price only affects for 1 second and if we clamp the rate of change per second that shouldn't matter much at all

xenide: Nope, because we only update the 1st trade of the timestamp, when they trade / arb back to the normal price in the same block, nothing happens to that oracle reading in the 2nd trade of the block. 

**In the same block**
1st trade - manipulate oracle, inaccurate instant prices written to oracle 
2nd trade - arb it back so no one else does it, oracle sample unchanged 

Attacker monitors the pair to make sure no other trades take place: that way the instant prices get time to accumulate. If someone comes in to trade, the instant prices get yanked towards fair price. The attacker can repeat the same 2-trade action to mess with the instant prices again, and continue to monitor. 

To the outside world monitoring prices, it didn't change much. But sitting in the oracle sample is the manipulated instant prices waiting to be written and accumulated. 

This test case demonstrates that perhaps updating instant prices at every trade is necessary.

OliverNChalk: Instant prices don't get written into the oracle but you mean because the oracle price accumulates weight until the next trade it can influence the oracle up the max trade amount?

OliverNChalk: Feels like cleanest solution is to just write the oracle every swap?

xenide: Yes the cleanest is to write the oracle every swap, and only updating the instant price if it's not the first swap of the block, which is the previous impl with the `else` branch. 

This is the only way to force any manipulators to put their capital at risk (increase the cost) when moving prices, as there is no guarantee they can arb it back in the next block. And if they arb it back in the next block, then the manipulation only lasted for 1 second

src/ReservoirPair.sol
```diff
@@ -148,15 +150,47 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
+
+            // both balance should never be zero, but necessary to check so we don't pass 0 values into arithmetic operations
+            // shortcut to calculate aBalance0 > 0 && aBalance1 > 0
+            if (aBalance0 * aBalance1 > 0) {
+                Observation storage lPrevious = _observations[sSlot0.index];
+                (uint256 lInstantRawPrice, int256 lLogInstantRawPrice) = _calcSpotAndLogPrice(aBalance0, aBalance1);
+
+                // a new sample is not written for the first mint
+                // shortcut to calculate lTimeElapsed > 0 && aReserve0 > 0 && aReserve1 > 0
+                if (lTimeElapsed * aReserve0 * aReserve1 > 0) {
```
xenide, line 172: just want to document this for posterity: 

I attempted to simplify this condition to only `lTimeElapsed > 0`, which therefore implies that an oracle sample is created in the very first mint. However, I did not adopt this approach in the end for one edge case. In the edge case of a swap happening after the very first mint (pair creation). the previous timestamp would not be zero, and therefore the clamped price would not be correctly calculated (it'd always be 0 cuz the time elapsed is 0). 

So it's back to the original approach of writing only the instant prices for the very first block (regardless of how many things happen in that block), and use the previous oracle observation as a "placeholder" to temporarily store the instant prices for accumulation in the future. This approach is more semantically correct as we're not "creating" a sample (as we don't write a timestamp), but instead just using it as a placeholder.

test/unit/ReservoirPair.t.sol
OliverNChalk: Why the need for an else if you write the update on every swap, because otherwise it will move the index?

xenide: yes, then we'll have multiple samples per timestamp which creates problems when reading the oracle.

The alternative is to read if `lPrevious.timestamp` is the current timestamp we're attempting to write, and if so, only overwrite the instant prices. But that's basically an `else` as well

src/ReservoirPair.sol
```diff
@@ -148,15 +152,47 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
+
+                    (, int256 lLogInstantClampedPrice) = _calcClampedPrice(
+                        lInstantRawPrice,
+                        lLogInstantRawPrice,
+                        LogCompression.fromLowResLog(lPrevious.logInstantClampedPrice),
+                        lTimeElapsed,
+                        lPrevious.timestamp
+                    );
+
+                    _updateOracleInstantPrices(lPrevious, lLogInstantRawPrice, lLogInstantClampedPrice);
```
OliverNChalk, line 195: Does this write to the oracle or slot0/cache? Cause with the other values we write them to cache and only update the oracle once per slot. Does this mean that we are writing the oracle every trade, is this fine/a problem - maybe not on L2?

src/ReservoirPair.sol
```diff
@@ -480,19 +516,23 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
     //////////////////////////////////////////////////////////////////////////*/
 
     event OracleCallerUpdated(address oldCaller, address newCaller);
-    event MaxChangeRateUpdated(uint256 oldMaxChangePerSecond, uint256 newMaxChangePerSecond);
+    event ClampParamsUpdated(uint128 newMaxChangeRatePerSecond, uint128 newMaxChangePerTrade);
 
     // 100 basis points per second which is 60% per minute
```
OliverNChalk: 100bp = 1 percent, so easier to just say 1% per second

src/ReservoirPair.sol
```diff
@@ -509,42 +549,93 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
+        uint256 aPreviousTimestamp
+    ) internal virtual returns (uint256 rClampedPrice, int256 rClampedLogPrice) {
         // call to `percentDelta` will revert if the difference between aCurrRawPrice and aPrevClampedPrice is
         // greater than uint196 (1e59). It is extremely unlikely that one trade can change the price by 1e59
-        if (aCurrRawPrice.percentDelta(aPrevClampedPrice) > maxChangeRate * aTimeElapsed) {
+        if (
+            (
+                aCurrRawPrice.percentDelta(aPrevClampedPrice) <= maxChangeRate * aTimeElapsed
+                    && aCurrRawPrice.percentDelta(aPrevClampedPrice) <= maxChangePerTrade
+            )
```
OliverNChalk: Can we extract these bools to variables and then do `condition0 && condition1`? Use better names than i have tho

src/ReservoirPair.sol

src/ReservoirPair.sol

src/ReservoirPair.sol
xenide: this only writes to the oracle observations. This case only happens for 2nd-and-later swaps in the same block.

OliverNChalk (approved):

PR #228, opened 2024-03-16
feat: introduce max change per trade

### LINKED ISSUE #226: Caveats regarding the clamp
Say on an L2 (e.g. Arbitrum), if the sequencer goes down for a few hours, no swap would be able to take place. As such, the first swap that happens can swing the instant clamped (and of course raw) price a lot. In other words, the clamp becomes quite useless if there are no swaps for an extended period of time, or put it inversely, the clamp works best when there are frequent swaps.

e.g. with a 1bp/s rate of change, if there are no swaps for 2 hours (7200 s), the swap can move the price by 72% and it would not exceed the clamp.

Potential mitigations: 
1. New borrows use the instant price in the case where there has been no oracle observation for X period 
2. Liquidations use a longer TWAP (e.g. 1-hr TWAP) instead of the normal TWAP (10-15 mins)
3. Clamp the max change within one trade
  - 5% ?

xenide:
## Solution

with the implementation of this extra check (max change per trade), there are 4 possible scenarios: 

1. raw price doesn't violate both conditions
  - simple: clampedPrice = rawPrice 
2. raw price only violates rate of change, doesn't violate max change per trade
  - use the rate-of-change capped price  
  - is it possible that the max-change-per-trade capped price is a smaller change than rate-of-change-capped price? 
    - e.g. maxChangePerTrade is 2%, rate of change is 3bp/s. 
      - trade comes after 30 seconds, doing a 1% change in price. 
      - max-change-per-trade capped price is 102%, rate-of-change capped price is 100.9%
    - e.g. maxChangePerTrade is 2%, rate of change is 100bp/s. 
      - trade comes after 3 seconds, doing a 1% change in price
      - max-change-per-trade capped price is 102%, rate-of-change capped price is 103%
    - seems that it's not possible
3. raw price only violates max change per trade, doesn't violate rate of change
  - use the max-change-per-trade capped price 
4. raw violates both rate of change and max change per trade
  - should we take the rate-of-change capped or max-change-per-trade capped? 
  - should choose the closer one? Arithmetic / geometric mean of the two?
    - Arithmetic mean would be cheaper than geometric mean due to gas cost 
  - Propose to choose the closer one (smaller change)

### DIFF
--- a/src/ReservoirPair.sol
+++ b/src/ReservoirPair.sol
@@ -49,7 +49,7 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
             updateSwapFee();
             updatePlatformFee();
             updateOracleCaller();
-            setMaxChangeRate(factory.read(MAX_CHANGE_RATE_NAME).toUint256());
+            setClampParams(factory.read(MAX_CHANGE_RATE_NAME).toUint128(), factory.read(MAX_CHANGE_PER_TRADE_NAME).toUint128());
         }
     }
 
@@ -514,18 +514,21 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
     //////////////////////////////////////////////////////////////////////////*/
 
     event OracleCallerUpdated(address oldCaller, address newCaller);
-    event MaxChangeRateUpdated(uint256 oldMaxChangePerSecond, uint256 newMaxChangePerSecond);
+    event ClampParamsUpdated(uint128 newMaxChangeRatePerSecond, uint128 newMaxChangePerTrade);
 
     // 100 basis points per second which is 60% per minute
     uint256 internal constant MAX_CHANGE_PER_SEC = 0.01e18;
     string internal constant MAX_CHANGE_RATE_NAME = "Shared::maxChangeRate";
+    string internal constant MAX_CHANGE_PER_TRADE_NAME = "Shared::maxChangePerTrade";
     string internal constant ORACLE_CALLER_NAME = "Shared::oracleCaller";
 
     mapping(uint256 => Observation) internal _observations;
 
     // maximum allowed rate of change of price per second to mitigate oracle manipulation attacks in the face of
     // post-merge ETH. 1e18 == 100%
-    uint256 public maxChangeRate;
+    uint128 public maxChangeRate;
+    // how much the clamped price can move within one trade. 1e18 == 100%
+    uint128 public maxChangePerTrade;
 
     address public oracleCaller;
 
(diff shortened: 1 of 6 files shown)


src/ReservoirPair.sol
```diff
@@ -557,19 +565,37 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
             // aTimeElapsed <= 32 bits
             if (aCurrRawPrice > aPrevClampedPrice) {
-                rClampedPrice = aPrevClampedPrice.fullMulDiv(1e18 + maxChangeRate * aTimeElapsed, 1e18);
+                uint256 lRateLimitedPrice = aPrevClampedPrice.fullMulDiv(1e18 + maxChangeRate * aTimeElapsed, 1e18);
+                uint256 lPerTradeLimitedPrice = aPrevClampedPrice.fullMulDiv(1e18 + maxChangePerTrade, 1e18);
+                rClampedPrice = lRateLimitedPrice.min(lPerTradeLimitedPrice);
+                assert(rClampedPrice < aCurrRawPrice);
             } else {
-                assert(aPrevClampedPrice > aCurrRawPrice);
-                rClampedPrice = aPrevClampedPrice.fullMulDiv(1e18 - maxChangeRate * aTimeElapsed, 1e18);
```
xenide, line 574: I think the auditors missed this. This underflow vuln has got to be a critical, and is live on our production pairs right now

src/ReservoirPair.sol
```diff
@@ -557,19 +565,37 @@ abstract contract ReservoirPair is IAssetManagedPair, ReservoirERC20 {
...
             (rClampedPrice, rClampedLogPrice) = (aCurrRawPrice, aCurrLogRawPrice);
         } else {
             // clamp the price
             // multiplication of maxChangeRate and aTimeElapsed would not overflow as
             // maxChangeRate <= 0.01e18 (50 bits)
             // aTimeElapsed <= 32 bits
             if (aCurrRawPrice > aPrevClampedPrice) {
-                rClampedPrice = aPrevClampedPrice.fullMulDiv(1e18 + maxChangeRate * aTimeElapsed, 1e18);
+                uint256 lRateLimitedPrice = aPrevClampedPrice.fullMulDiv(1e18 + maxChangeRate * aTimeElapsed, 1e18);
+                uint256 lPerTradeLimitedPrice = aPrevClampedPrice.fullMulDiv(1e18 + maxChangePerTrade, 1e18);
```
xenide: I think an optimization we can do here, is to first calculate the lower rate of change, and then do only one `fullMulDiv`, as a `fullMulDiv` is expensive

OliverNChalk: You don't need to clamp the output, you can just clamp the multiplier. I.e. `min(maxChangeRate * aTimeElapsed, maxChangePerTrade)`

src/ReservoirPair.sol
OliverNChalk: What's the vuln, it freezes the pair if it overflows? How realistic is 1e18 seconds elapsed, surely that's like super far away?

xenide: it's the product of `maxChangeRate` and `aTimeElapsed` that's easy to overflow `1e18`. 

e.g. 86400 (seconds in a day) * 0.00003e18 (0.3bp / s)  = 2.592e18 > 1e18

Yes it freezes the pair as it's an arithmetic underflow. I discovered it during regression testing. Kinda weird that it didn't show up before

src/ReservoirPair.sol
xenide: yup totally. Implemented

PR #231, opened 2024-11-17
feat: integrate with Euler V2 vaults

xenide:
## Motivation

We need to integrate our AMM with Euler instead of AAVE, since we will be supplying our assets there. 

### Outstanding items:
- [x] handle pause / frozen cases for euler 
   - not handling, in line with how we integrated with aave previously
- [x] integrate additional rewards 
- [x] track total shares and only allow the admin to set another vault when the totalShares is 0


test/integration/Euler.t.sol
```diff
@@ -13,7 +13,7 @@ import { IRewardsController } from "src/interfaces/aave/IRewardsController.sol";
 
 import { FactoryStoreLib } from "src/libraries/FactoryStore.sol";
```
OliverNChalk, line 9: why so many lines deleted in tests, do we still have sufficient coverage?

src/asset-management/EulerV2Manager.sol
```diff
@@ -305,17 +274,50 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
+            lSharesAllocated += lNewSharesEntitled;
+        }
 
-        rClaimed = rewardsController.claimRewards(lMarkets, type(uint256).max, msg.sender, aReward);
+        // the last in the list will take all the remaining shares, and sometimes will get 1 or 2 more than they're entitled to
+        // due to the rounding down in previous calculations for other pairs
+        // this is to prevent the sum of each pair+token's shares not summing up to totalShares
+        shares[aPairs[lLength - 1]][aAsset] = lNewShares - lSharesAllocated;
+        lSharesAllocated += shares[aPairs[lLength - 1]][aAsset];
+        require(lSharesAllocated == lNewShares, "AM: REWARD_SHARES_MISMATCH");
```
OliverNChalk, line 319: Why not just have the interface be a single pair, what value do we get from this multi pair support (cause we take a lot of risk, for instance this could fail if we get unlucky with rounding).

src/asset-management/EulerV2Manager.sol
```diff
@@ -128,62 +108,46 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
         private
         returns (uint256 rShares)
     {
-        uint256 lTotalShares = totalShares[aAaveToken];
-        if (totalShares[aAaveToken] == 0) {
-            rShares = aAmount;
-        } else {
-            rShares = aAmount * lTotalShares / aAaveToken.balanceOf(address(this));
-        }
+        rShares = aVault.previewDeposit(aAmount);
```
OliverNChalk: What is previewDeposit, does it actually deposit?

src/asset-management/EulerV2Manager.sol
xenide: it's because from the perspective of the vaults, the asset manager is the owner of all these shares, and when we claim the rewards, we should claim everything at one go. The approach of doing it for every single pair would work too but it's more txs and calculations.

actually, this math guarantees that this operation won't fail. 

In the past with AAVE it was simpler cuz the guardians (we) could just dump whatever reward token, convert it to aaveTokens, and transfer it back to the asset manager, and the shares proportion would take care of itself.

src/asset-management/EulerV2Manager.sol
xenide: this function simulates a deposit, and tells us how many shares it returns but doesn't actually deposit the assets. Honestly we can just skip this step, and do the deposit right away, and keep track of the shares from the return value of `deposit()`

test/integration/Euler.t.sol
xenide: the deleted cases are for cases in AAVE where the pool is frozen or paused. For Euler, the pools will never be these states so the test cases are no longer applicable. 

Good idea for coverage, I should add it, don't think we're doing coverage on this at the moment.

src/asset-management/EulerV2Manager.sol
OliverNChalk: Sounds good

src/asset-management/EulerV2Manager.sol
```diff
@@ -10,63 +10,51 @@ import { Address } from "@openzeppelin/utils/Address.sol";
...
     event Investment(IAssetManagedPair pair, IERC20 token, uint256 shares);
     event Divestment(IAssetManagedPair pair, IERC20 token, uint256 shares);
 
-    /// @dev tracks how many aToken each pair+token owns
+    error OutstandingSharesForVault();
+
+    /// @dev Mapping from an ERC20 token to an Euler V2 vault.
+    /// This implies that for a given asset, there can only be one vault.
+    /// If the admin of the manager wishes to specify a different vault for an asset, they would have to manually ensure that all pairs have
+    /// divested, otherwise the pairs might not be able to retrieve their assets.
```
OliverNChalk: These lines should wrap at the same width (80 or 100 col)

src/asset-management/EulerV2Manager.sol
```diff
@@ -305,17 +263,61 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
+        bytes32[][] calldata aProofs
+    ) external onlyGuardianOrOwner {
+        aDistributor.claim(aUsers, aTokens, aAmounts, aProofs);
+
+        for (uint256 i = 0; i < aTokens.length; ++i) {
+            SafeTransferLib.safeTransfer(
+                aTokens[i],
+                msg.sender,
+                // the amounts specified in the argument might not be the actual amounts disimbursed by the distributor, due to the possibility of having done the claim previously
+                // thus it is necessary to use balanceOf to transfer the correct amount
```
OliverNChalk: excessively long comment line, limit to 80 or 100

src/asset-management/EulerV2Manager.sol
```diff
@@ -305,17 +263,61 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
     //////////////////////////////////////////////////////////////////////////*/
 
-    function claimRewardForMarket(address aMarket, address aReward)
+    function claimRewards(
+        IDistributor aDistributor,
+        address[] calldata aUsers,
+        address[] calldata aTokens,
+        uint256[] calldata aAmounts,
+        bytes32[][] calldata aProofs
+    ) external onlyGuardianOrOwner {
```
OliverNChalk, line 267: This is the interface for euler, right? If we need to change we can just do windown mode and migrate?

src/asset-management/EulerV2Manager.sol
xenide: okay fixed. Also made forge fmt format the comments as well. Previously it was excluded. However, line_length remains at the default at 120 for now

src/asset-management/EulerV2Manager.sol
xenide: fixed as well

src/asset-management/EulerV2Manager.sol
xenide: this is the interface for merkl, a rewards distribution protocol which euler and its partners have chosen to distribute additional incentives (on top of the base yield). think of it as minichef as a service, fuss free. 

Yes exactly when needed, we can always wind down and migrate. If there rewards not claimable by this function, we can always use the `rawCall` function

OliverNChalk: Perfect sounds good

test/integration/Euler.t.sol
xenide: achieved 100% of the asset manager code: 

https://app.codecov.io/gh/reservoir-labs/amm-core/blob/feat%2Fsupply-euler-vaults/src%2Fasset-management%2FEulerV2Manager.sol

src/asset-management/EulerV2Manager.sol
```diff
@@ -198,14 +152,14 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
+        IERC4626 lToken1Vault = assetVault[lToken1];
 
-        // do not do anything if there isn't a market for the token
-        if (address(lToken0AToken) == address(0)) {
+        // do not do anything if there isn't a designated vault for the token
+        if (address(lToken0Vault) == address(0)) {
             aAmount0Change = 0;
         }
-        if (address(lToken1AToken) == address(0)) {
+        if (address(lToken1Vault) == address(0)) {
```
OliverNChalk, line 159: I checked and it looks like it is but just wanna confirm that setting the delta to zero as there is no change is safe for the callers? For `adjustManagement` and `returnAsset` we could just use `require` right? Then if we made `_calculateChangeAmount` set amount to 0 when there is no manager that wouldnt need this as well?

I guess then we would have two map lookups (though on a warm slot), not sure how bad that is.

What do you think about require vs modifying argument to 0.

src/asset-management/EulerV2Manager.sol
```diff
@@ -305,17 +261,64 @@ contract AaveManager is IAssetManager, Owned(msg.sender), ReentrancyGuard {
...
+    /// @dev The guardian or owner would first call `claimRewards` and sell it for the underlying token.
+    /// The asset manager pulls the assets, deposits it to the vault, and distribute the proceeds in the form of ERC4626
+    /// shares to the pairs.
+    /// Due to integer arithmetic the last pair of the array will get one or two more shares, so as to maintain the
+    /// invariant that the sum of shares for all pair+token equals the totalShares.
+    function distributeRewardForPairs(IERC20 aAsset, uint256 aAmount, IAssetManagedPair[] calldata aPairs)
         external
         onlyGuardianOrOwner
-        returns (uint256 rClaimed)
+        nonReentrant
```
OliverNChalk, line 290: This is the only function marked non reentrant why does this need it and not the others?

OliverNChalk (commented):
Some small comments so far. Really good looking code, hard to find things to complain about haha

src/asset-management/EulerV2Manager.sol
xenide: in the current impl, setting those amounts to 0 is safe for the callers -- the caller won't get a revert. 

Long time ago when I was designing this, I had this consideration too. 

So I think I decided that validating this in `_adjustManagement` is more correct, because it's checking where it matters (here in the core function), right before it interacts with the vaults. If we do it in `adjustManagement` and `returnAsset` it's also correct but it's doing it in the periphery, so to speak. And one day if we introduce another function but forget to check for these conditions it could be problematic. 

In other words, to be robust, we need to validate the inputs in the heart, i.e. `_adjustManagement`. If we do it in the external functions, I think it's at best duplicating the work and at worst opening up to vulnerabilities.

src/asset-management/EulerV2Manager.sol
xenide: actually it's this function plus the internal `_adjustManagement` that have the reentrancy guard. 

The reason is that only these two function change the state, so we have to be take more precautions. 

The rest the external functions are either view or are admin functions that don't modify the important states (`shares` and `totalShares`)

src/asset-management/EulerV2Manager.sol
OliverNChalk: Right but why not just revert in _adjustManagement?

src/asset-management/EulerV2Manager.sol
OliverNChalk: Sounds good, nice

src/asset-management/EulerV2Manager.sol
xenide: cuz `_calculateChangeAmount` used by `afterLiquidityEvent` has no knowledge of whether a certain token has an assigned vault or not. 

And I rather let it do some extra math than do an extra sload to check for the vault. Granted, this sload slot will be warm, but still at least 100 gas. I haven't measured it but my guesstimation is that a few arithmetic operations should still be < 100 gas

OliverNChalk: Sounds good

OliverNChalk (approved):

PR #233, opened 2024-11-28
feat: use transient storage for reentrancy guard

### LINKED ISSUE #230: fix: consider using newly introduced transient storage for reentrancy guard
solidity 0.8.24 has introduced `TSTORE` and `TLOAD` opcodes. This should be cheaper to use as a reentrancy guard compared to the normal SSTORE + SLOAD. 

See https://github.com/Vectorized/soledge/blob/main/src/utils/ReentrancyGuard.sol

xenide:


OliverNChalk (approved):
Looks good, like the simplification

PR #234, opened 2024-12-02
feat: fold `StableMintBurn` back into `StablePair`

xenide:
## Motivation

- thanks to the removal of some functionality, custom errors, and `via-ir`, we could achieve cutting the bytecode of StablePair down to under the limit


src/curve/constant-product/ConstantProductPair.sol
```diff
@@ -101,7 +105,7 @@ contract ConstantProductPair is ReservoirPair {
         uint256 lAmount1 = lBalance1 - lReserve1;
 
         _mintFee(lReserve0, lReserve1);
-        uint256 lTotalSupply = totalSupply(); // gas savings, must be defined here since totalSupply can update in _mintFee
+        uint256 lTotalSupply = totalSupply(); // gas savings, must be defined here since totalSupply can update in // _mintFee
```
OliverNChalk: remove the `//`

OliverNChalk (approved):
minor comment fix

src/curve/constant-product/ConstantProductPair.sol
xenide: fixed

PR #239, opened 2024-12-10
test: stable oracle math

xenide:
## Motivation

Wanted to: 

1. verify that the balancer's impl of calculating spot price is correct, since there was no derivation provided for the formula. Attempted to look at other derivations such as [this](https://x.com/0xa9a/status/1514192791689179137). But despite my efforts at algebraic manipulation but it's not evident that they are equivalent. 
2. ensure that the arithmetic operations in `calcSpotPrice` will never revert even in the most unlikely circumstances

## Solution

1. use a very small amount of x to simulate `dx` in the derivative sense. And with that we have shown that balancer's formula for calculating spot price is correct. 
2. introduced a few safeguards to ensure that the math will never cause an underflow


OliverNChalk (approved):
unable to review the math in depth but looks reasonable

